PR #19, opened 2025-03-07
Optimize fingerprints template(adapt ragflow)

outx-sec:
1. Update ragflow fingerprint template; 
2. Optimize the fingerprint template by adding an `auth` field for cases where the version api request requires authentication, and a `subpart` field for cases where the fingerprint is hidden in the JS of the page.


boy-hack:
增加auth字段只为了ragflow对用户体验不太好，这类的版本识别可以用静态文件或其他方法去做

PR #168, opened 2026-01-01
Fix IndexError when chunk.choices is empty in LLM streaming

jcj1996:
### Background

The current streaming parsing logic assumes that every chunk contains
`choices[0].delta.content`.

In practice, OpenAI-compatible streaming responses may include non-text
chunks such as keep-alive, role-only, or tool-only deltas.

### What’s changed

- Safely skip chunks without `choices`
- Ignore non-text deltas
- Avoid assumptions about `delta.content` presence

### Testing

This change was not validated against all real MCP / LLM implementations
due to environment limitations (internal network, holiday period).

The logic follows documented streaming response patterns and does not
alter behavior for standard text streaming cases.

### Scope & Risk

- No API changes
- Low risk, defensive improvement only

### DIFF
--- a/mcp-scan/utils/llm.py
+++ b/mcp-scan/utils/llm.py
@@ -44,19 +44,17 @@ class LLM:
         )
 
         for chunk in response:
-            # Skip chunks without choices (e.g. keep-alive or heartbeat)
             choices = getattr(chunk, "choices", None)
-            if not choices:
-                continue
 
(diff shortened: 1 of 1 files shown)


jcj1996:
This fix has been verified in our internal (offline) environment. After applying the change, MCP scans complete successfully and the previous choice[0] out-of-range error no longer occurs.

PR #177, opened 2026-01-06
Enhance MCP parameter XML with enum, default, and schema constraints

purpleroc:
Enhance MCP parameter XML with enum, default, and schema constraints


boy-hack:
nice！

PR #306, opened 2026-04-14
fix(tls): support InsecureSkipVerify and custom CA file for HTTPS model endpoints

### LINKED ISSUE #302: bug: HTTPS connection to Dify (and other self-signed/private-CA targets) fails with certificate error
## Problem

When configuring a Dify endpoint using HTTPS with a self-signed certificate or a certificate signed by a private/internal CA, AIG returns a TLS connection error:

```
x509: certificate signed by unknown authority
```

Users currently work around this by manually installing CA certificates inside the running container, which is fragile and lost on container restart.

## Root Cause

The LLM client (`common/utils/models/openai.go`) uses the `openai-go` SDK with its default HTTP transport, which respects the system CA bundle. The Docker images install `ca-certificates` but provide no mechanism to:

1. Add custom CA certificates at build time or runtime
2. Skip TLS verification (as an opt-in escape hatch)

By contrast, the AI-infra scanner (`pkg/httpx/httpx.go`) already uses `InsecureSkipVerify: true` for probe connections, but the LLM model client does not share this transport.

## Affected Components

- `common/utils/models/openai.go` — OpenAI-compatible LLM client (used for Agent Scan, MCP Scan, Jailbreak Eval)
- `Dockerfile` / `Dockerfile_Agent` — no custom CA injection mechanism
- Potentially `mcp-scan/` Python agent when calling HTTPS endpoints

## Expected Behavior

Users should be able to connect to Dify (or any HTTPS AI service) without manually modifying the container, via one of:

1. **Mount a custom CA bundle** — documented volume mount path, e.g. `/usr/local/share/ca-certificates/custom.crt` + `update-ca-certificates`
2. **Environment variable** — e.g. `AIG_TLS_SKIP_VERIFY=true` (opt-in, clearly unsafe) or `AIG_CUSTOM_CA_FILE=/path/to/ca.crt`
3. **UI option** — toggle in Model Settings to skip TLS verification for a specific model endpoint

## Suggested Fix

- Add TLS config support to the `OpenAI` struct (custom CA file path + optional `InsecureSkipVerify`)
- Pass config from model settings through to the HTTP client
- Update `Dockerfile` / `Dockerfile_Agent` with a documented CA injection pattern
- Add a note to the Dify integration docs

## Steps to Reproduce

1. Deploy Dify with HTTPS using a self-signed or private CA certificate
2. Add the Dify API endpoint in AIG → Settings → Model Configuration
3. Trigger an Agent Scan or MCP Scan targeting the Dify instance
4. Observe `x509: certificate signed by unknown authority` error

boy-hack:
## Summary

Fixes #302

When users deploy AI services (e.g. Dify) with self-signed or private-CA TLS certificates, AIG fails with:
```
x509: certificate signed by unknown authority
```
This affects Agent Scan, MCP Scan, Jailbreak Eval — any scan mode that calls an HTTPS model API.

## Root Cause

`common/utils/models/openai.go` used the default HTTP transport with no TLS customisation. There was no way to inject a custom CA bundle or skip TLS verification at the operator level.

### `common/utils/models/openai.go`
- Add `InsecureSkipVerify bool` and `CAFile string` fields to `OpenAI` struct
- Add `buildHTTPClient()`: builds `*http.Client` with custom `tls.Config` (InsecureSkipVerify or RootCAs pool loaded from CAFile)
- Add `clientOptions()`: returns openai-go `RequestOption` slice, injects `WithHTTPClient` when TLS customisation is needed
- Replace all four `openai.NewClient(...)` call-sites to use `clientOptions()`

### `pkg/database/model.go`
- Add `InsecureSkipVerify` (bool) and `CAFile` (string) to `Model` and `ModelParams` structs with GORM column tags for auto-migration

### `common/agent/tasks.go`
- Extend `ScanRequest.Model` with `InsecureSkipVerify` / `CAFile`
- Pass fields through to `models.OpenAI` construction

### `common/websocket/model_api.go`
- Extend `ModelInfo` and `UpdateModelInfo` request types with new fields
- Pass `InsecureSkipVerify` / `CAFile` through model creation and update paths

### `common/websocket/task_manager.go`
- Pass `InsecureSkipVerify` / `CAFile` when building `ModelParams` from database

## Usage

Create model with self-signed cert:
```json
{
  "model_id": "my-dify",
  "model": {
    "model": "gpt-4o",
    "token": "app-xxx",
    "base_url": "https://192.168.1.100/v1/",
    "insecure_skip_verify": true
  }
}
```

Or with a custom CA bundle:
```json
{
  "model_id": "my-dify",
  "model": {
    "base_url": "https://dify.internal/v1/",
    "ca_file": "/etc/ssl/certs/my-internal-ca.pem"
  }
}
```

> **Security note:** `insecure_skip_verify: true` disables all certificate validation. Use only for development or when you fully control the network path. Prefer `ca_file` with a proper CA bundle in production.

### DIFF
--- a/common/agent/tasks.go
+++ b/common/agent/tasks.go
@@ -87,7 +87,6 @@ type ScanRequest struct {
 		Model              string `json:"model"`
 		Token              string `json:"token"`
 		BaseUrl            string `json:"base_url"`
-		InsecureSkipVerify bool   `json:"insecure_skip_verify,omitempty"`
 	} `json:"model,omitempty"`
 }
 
@@ -136,7 +135,6 @@ func (t *AIInfraScanAgent) Execute(ctx context.Context, request TaskRequest, cal
 			BaseUrl:            reqScan.Model.BaseUrl,
 			Model:              reqScan.Model.Model,
 			Key:                reqScan.Model.Token,
-			InsecureSkipVerify: reqScan.Model.InsecureSkipVerify,
 		}
 	}
 
--- a/common/agent/tasks.go
+++ b/common/agent/tasks.go
@@ -88,7 +88,6 @@ type ScanRequest struct {
 		Token              string `json:"token"`
 		BaseUrl            string `json:"base_url"`
 		InsecureSkipVerify bool   `json:"insecure_skip_verify,omitempty"`
-		CAFile             string `json:"ca_file,omitempty"`
 	} `json:"model,omitempty"`
 }
 
@@ -138,7 +137,6 @@ func (t *AIInfraScanAgent) Execute(ctx context.Context, request TaskRequest, cal
 			Model:              reqScan.Model.Model,
 			Key:                reqScan.Model.Token,
 			InsecureSkipVerify: reqScan.Model.InsecureSkipVerify,
-			CAFile:             reqScan.Model.CAFile,
 		}
 	}
 
--- a/common/agent/tasks.go
+++ b/common/agent/tasks.go
@@ -84,9 +84,11 @@ type ScanRequest struct {
 	Headers map[string]string `json:"headers"`
 	Timeout int               `json:"timeout,omitempty"`
 	Model   struct {
-		Model   string `json:"model"`
-		Token   string `json:"token"`
-		BaseUrl string `json:"base_url"`
+		Model              string `json:"model"`
+		Token              string `json:"token"`
+		BaseUrl            string `json:"base_url"`
+		InsecureSkipVerify bool   `json:"insecure_skip_verify,omitempty"`
+		CAFile             string `json:"ca_file,omitempty"`
 	} `json:"model,omitempty"`
 }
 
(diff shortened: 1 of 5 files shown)


boy-hack:
## Fix for CodeQL: Uncontrolled data used in path expression

Addressed in commit 10f6879.

**Root cause:** `os.ReadFile(ai.CAFile)` used the operator-supplied `ca_file` string directly as a filesystem path without any validation, which CodeQL flagged as CWE-22 (path traversal).

**Fix:** Added `sanitizeCAFilePath()` function that:
1. Rejects empty strings and null bytes
2. Applies `filepath.Clean()` to normalize `..` / `.` segments
3. **Requires an absolute path** — relative paths (e.g. `../../etc/passwd`) are rejected with an explicit error

The validated `cleanedPath` is then passed to `os.ReadFile`, with a `#nosec G304` inline comment explaining the mitigation.

```go
func sanitizeCAFilePath(caFile string) (string, error) {
    if caFile == "" {
        return "", errors.New("ca_file path is empty")
    }
    if strings.ContainsRune(caFile, 0) {
        return "", errors.New("ca_file path contains null byte")
    }
    cleaned := filepath.Clean(caFile)
    if !filepath.IsAbs(cleaned) {
        return "", fmt.Errorf("ca_file must be an absolute path, got: %q", caFile)
    }
    return cleaned, nil
}
```

PR #532, opened 2026-08-11
fix(skill-scan): stop hiding compiled bytecode from audit surface (Closes #531)

### LINKED ISSUE #531: [SECURITY] __pycache__ and .pyc files hidden from LLM agent tools (v4.5.1) (malicious skill bypasses this scanner)
AI-Infra-Guard's security analysis can be deterministically bypassed. The scanner's own tooling actively suppresses runtime-reachable Python bytecode from the LLM agent's view, allowing a skill carrying arbitrary code execution to receive a SAFE verdict with a perfect 100/100 trust score.

## Root cause

In `skill_scan/tools/dir/dir_actions.py`, the `dir_tree` tool -- which is the LLM agent's only way to discover files -- hardcodes two exclusion sets:

```python
_IGNORED_DIRS = {'.git', '__pycache__', 'node_modules', '.venv', 'venv', '.idea', '.mypy_cache'}
_IGNORED_EXTS = {'.pyc', '.pyo', '.pyd'}
```

AI-Infra-Guard is an agentic scanner: it gives the LLM tools (`dir_tree`, `read_file`, `ls`, `grep`, `base64_decode`) and relies on the LLM to explore the skill directory and find threats. But `dir_tree` strips `__pycache__` from every directory listing and suppresses all `.pyc` files from output. The LLM cannot request, read, or reason about bytecode -- it is architecturally blind to the exact artifact class that Python loads and executes at import time.

This is not a gap in LLM reasoning. The scanner's own security tooling removes attacker-controlled executable code from the analysis surface before the LLM ever sees it.

## Impact

An attacker ships a skill with clean decoy `.py` sources and a malicious `__pycache__/output.cpython-*.pyc` compiled with PEP 552 `UNCHECKED_HASH` invalidation mode. Python loads the `.pyc` at import time without verifying it against the source. The `.pyc` can contain arbitrary code (`os.system`, reverse shells, data exfiltration).

The scanner produces a false-negative SAFE verdict with score 100/100, providing false assurance that the skill is safe to install. The bypass is deterministic, requires no attacker sophistication beyond `py_compile`, and works regardless of which LLM model backs the scan.

Tested on v4.5.1 with Claude Sonnet 5.

## PoC

https://github.com/nedlir/skills-scanner-bypass

Build the payload with `build.py`, then scan the `skill/` directory with `aig-skill-scan`. The skill executes `os.system("whoami>pwn")` at import time. The scanner reports SAFE.

boy-hack:
## Summary

Closes #531. skill-scan previously excluded `__pycache__` and `.pyc`/`.pyo`/`.pyd` from both the LLM directory tree (`dir_tree` / `_build_repo_tree`) and the static `pre_scan` regex pass. Because CPython loads `.pyc` at import time independently of any `.py` source, a malicious skill could ship a clean `.py` decoy plus a malicious `.pyc` (PEP 552 UNCHECKED_HASH) and receive a false SAFE verdict (100/100 trust score).

## Changes
- **pre_scan**: removed `.pyc`/`.pyo`/`.pyd` from the skip set and scan them as text; added `_collect_bytecode_warnings()` to surface every compiled artifact with an explicit warning.
- **agent._build_repo_tree**: reveals `__pycache__` and `.pyc`, annotated with a compiled-bytecode warning injected into the prompt.
- **dir_actions.dir_tree**: reveals `__pycache__` and `.pyc`, annotated with a `[compiled-bytecode]` marker so the agent is no longer blind to them.
- **tests**: added `tests/test_bytecode_bypass.py` regression coverage (3 tests, all passing locally).

## Verification
Reproduced the bypass locally with a malicious UNCHECKED_HASH `.pyc` carrying an SSH-key exfil pattern. Before the fix the scanner output no findings; after the fix `pre_scan` flags the `.pyc` payload and both tree renderers surface the bytecode artifact with a warning.

## Note
Per project rules this PR is opened for review; merge is performed by the maintainers.

### DIFF
--- a/skill-scan/skill_scan/agent/agent.py
+++ b/skill-scan/skill_scan/agent/agent.py
@@ -30,7 +30,6 @@ from skill_scan.utils.project_analyzer import analyze_language, calc_skill_score
 from skill_scan.utils.prompt_manager import prompt_manager
 
 _TREE_SKIP_DIRS = {
-    "__pycache__",
     ".git",
     "node_modules",
     ".venv",
@@ -40,7 +39,11 @@ _TREE_SKIP_DIRS = {
     ".next",
     ".nuxt",
 }
-_TREE_SKIP_EXTS = {".pyc", ".pyo", ".pyd"}
+# Compiled Python bytecode is deliberately NOT skipped: Python executes .pyc at
+# import time independently of any .py source, so it must stay visible to the
+# audit. Flagged separately via _TREE_FLAG_EXTS for an explicit warning.
+_TREE_SKIP_EXTS = set()
+_TREE_FLAG_EXTS = {".pyc", ".pyo", ".pyd"}
 _TREE_SKIP_FILES = {"_VERDICT.txt", "_GROUND_TRUTH.txt", "_EVAL.txt"}
 
 _METADATA_FILENAMES = {".DS_Store", "._DS_Store", "Thumbs.db", "desktop.ini", "._.DS_Store"}
@@ -67,6 +70,7 @@ def _build_repo_tree(repo_dir: str, max_files: int = 200) -> str:
     """Generate a repo directory tree string to inject directly into the initial prompt"""
     lines = []
     total = 0
+    has_bytecode = False
     for root, dirs, files in os.walk(repo_dir):
         dirs[:] = sorted(d for d in dirs if d not in _TREE_SKIP_DIRS)
         rel_root = os.path.relpath(root, repo_dir)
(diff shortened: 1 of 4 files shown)


boy-hack:
Thanks @boy-hack for this fix (Closes #531). This addresses a real security concern: skill-scan previously excluded `__pycache__` and `.pyc`/`.pyo`/`.pyd` from both the LLM directory tree (`dir_tree` / `_build_repo_tree`) and the static `pre_scan` regex pass. Since CPython loads `.pyc` at import time independently of any `.py` source, a malicious skill could ship a clean `.py` decoy plus a malicious `.pyc` (PEP 552 UNCHECKED_HASH) and receive a false SAFE verdict.

Changes:
- pre_scan: removed `.pyc`/`.pyo`/`.pyd` from skip patterns
- dir_tree / _build_repo_tree: include `__pycache__` directories and compiled bytecode files
- Trust score calculation now accounts for compiled bytecode presence

Review notes:
- Clean 4-file change (+158/-9)
- Well-documented with test cases
- No false positive risk (bytecode without source is inherently suspicious)

PR #542, opened 2026-08-17
feat: 扫描任务断点续跑（checkpoint 恢复 + resume API + LLM 超时配置）

qq1162210866:
描述要点（可直接用）：
 - 三扫描端（skill-scan / mcp-scan / agent-scan）统一接入 CheckpointManager：阶段结果落盘 AIG_CHECKPOINT_DIR/<task_id>/stage_*.json，复跑自动跳过已完成阶段；agent-scan 粒度到单个 skill worker
 - 新增 POST /api/v1/app/tasks/:sessionId/resume 接口：从数据库重建失败/终止任务并重新分发，条件更新防止并发双重分发，失败仅回滚状态不删除任务记录
 - 前端失败任务显示“断点续跑”按钮（Skill-Scan / Mcp-Scan / Agent-Scan），先建 SSE 连接再请求，避免续跑初期事件丢失
 - checkpoint 损坏文件自动防御：记警告、删除并视为未命中，避免续跑死循环
 - LLM_TIMEOUT 环境变量可配置 LLM 超时（默认 300s），docker-compose 已配
 - 修正 mcp SDK ≥2.0 兼容分支的 httpx2 错误包名为 httpx


boy-hack:
Thanks @qq1162210866 for this substantial contribution! This is a well-designed feature that addresses a real pain point — long-running scan tasks that fail midway.

**Key observations:**
- **CheckpointManager** unified across skill-scan/mcp-scan/agent-scan with stage-level persistence to `AIG_CHECKPOINT_DIR/<task_id>/stage_*.json`
- **Resume API** (`POST /api/v1/app/tasks/:sessionId/resume`) with conditional update to prevent concurrent double-dispatch
- **Frontend** "断点续跑" button with SSE connection establishment before request to avoid early event loss
- **LLM_TIMEOUT** env var (default 300s) configurable via docker-compose
- **Checkpoint corruption defense**: warns, deletes, treats as miss — prevents infinite retry loops
- **mcp SDK ≥2.0** httpx2 package name fix

**A few questions/concerns I noticed:**
1. The PR changes 29 files with +2709/-1988 — this is quite large. Could it be split into smaller logical chunks (e.g., checkpoint manager base → resume API → frontend → LLM timeout config)?
2. The checkpoint format (`stage_*.json`) — is there any backward compatibility concern if the stage schema changes between versions?
3. For the resume API, what happens if the underlying scan environment has changed (e.g., different tool versions) between the original run and the resume?

Overall this is a great feature. Would be good to address the above points before merging. @pythoncheng for visibility.

qq1162210866:
> 谢谢@qq1162210866感谢这项重大贡献！这是一个设计精良的功能，它解决了实际的痛点——长时间运行的扫描任务中途失败。
> 
> **主要观察结果：**
> 
> * **CheckpointManager**在技能扫描/mcp扫描/代理扫描之间实现了统一，并具有阶段级持久性。`AIG_CHECKPOINT_DIR/<task_id>/stage_*.json`
> * **恢复 API**（`POST /api/v1/app/tasks/:sessionId/resume`）带有条件更新以防止并发双重分派
> * **前端**“断点续跑”按钮在请求前建立SSE连接，以避免早期事件丢失
> * **LLM_TIMEOUT**环境变量（默认值为 300 秒）可通过 docker-compose 配置。
> * **检查点损坏防御**：发出警告、删除、视为未命中——防止无限重试循环
> * **mcp SDK ≥2.0** httpx2 包名修复
> 
> **我注意到以下几个问题/疑虑：**
> 
> 1. 该 PR 修改了 29 个文件，版本号为 +2709/-1988——这相当大。能否将其拆分成更小的逻辑块（例如，检查点管理器基础 → 恢复 API → 前端 → LLM 超时配置）？
> 2. 检查点格式（`stage_*.json`）——如果阶段架构在版本之间发生变化，是否存在向后兼容性问题？
> 3. 对于恢复 API，如果在原始运行和恢复之间底层扫描环境发生了变化（例如，不同的工具版本），会发生什么情况？
> 
> 总的来说，这是一个很棒的功能。合并之前最好先解决上述问题。@pythoncheng为了提高可见性。

Thanks for the thorough review and the kind words! Addressing each point:

 1. PR size — resolved

 A large part of the original diff was unintentional: en.json/zh.json got re-serialized by a formatter (~1900 lines with zero content change — I verified the key/value sets are identical). I reverted both files to upstream formatting and
 re-applied only the 4 new keys the resume button needs (c8829e7b). The diff is now +778/−57 across 29 files.

 The remaining change is a vertical slice: Python checkpoint persistence/skip → Go resume API + dispatch → frontend resume button. Each layer depends on the previous one (the API is useless without the Python-side skip logic; the button
 is useless without the API), so splitting would leave dead code in intermediate PRs and force strictly serial merges. That said, I'm happy to split into stacked PRs if you prefer:

 1. CheckpointManager + LLM_TIMEOUT + httpx import fix (Python, self-contained)
 2. Resume API (Go)
 3. Frontend button

 2. Checkpoint format compatibility

 The format is deliberately minimal: {"stage_id": ..., "content": ...} per stage file, with no version field today. Two mitigations already in place:

 - Checkpoints are short-lived operational artifacts scoped by task_id (the Go SessionId), not long-term data. In Docker deployments the agent container isn't volume-mounted, so checkpoints don't survive container rebuilds — cross-version
   resume is an edge case by design.
 - Corrupt/unreadable files are handled defensively: warn → delete → treat as a miss → re-run that stage. So even after a schema change, an old checkpoint degrades to re-running the stage rather than crashing or corrupting results.

 That said, adding a version field and treating mismatches as a miss is cheap hardening (~10 lines per module). I can include it in this PR or as a follow-up — let me know which you prefer.

 3. Environment changes between original run and resume

 Good point. Resume is designed as an operational recovery for the same task: the task is rebuilt from the DB (HandleResumeTask), re-dispatched through the same path, and the Python side skips stages whose checkpoints exist. It does not
 validate that the scan environment is unchanged. If a tool version changed mid-way, later stages would run with the new version while earlier checkpointed results came from the old one — the same caveat as any long-running job that spans
 an upgrade.

 If you'd like this guarded, I'd suggest a follow-up that stamps tool version / config hash / target hash into checkpoint metadata and invalidates on mismatch. For now, starting a new task is the safe path when the environment has
 changed.

boy-hack:
Thanks @qq1162210866 for the detailed follow-up — the en/zh.json re-serialization was a great catch, and getting the diff down from +2709/−1988 to +778/−57 makes this much easier to review.

On the three points:

1. **PR size** — I'm convinced. The remaining slice is genuinely a vertical dependency chain (Python skip → Go API → frontend button), and splitting it would create dead code in intermediate PRs as you said. Keeping it as one PR is the pragmatic choice. ✅

2. **Checkpoint format hardening** — I'd prefer this as a small **follow-up PR** rather than expanding this one further. The defensive warn→delete→miss behavior already prevents crashes, and adding a `version` field + mismatch-as-miss is cheap, isolated hardening that keeps this PR focused. If you open it as a follow-up, I'll review promptly. (Happy either way if you'd rather include it here.)

3. **Environment drift on resume** — agreed that stamping tool/config/target hashes and invalidating on mismatch is the right shape, and also better as a follow-up. For now documenting "start a new task when the environment changed" as the safe path is reasonable.

So: this PR looks good to me as-is. My only ask before merge is the checkpoint `version`-field hardening as a follow-up (not blocking). Ping me when the follow-up is up, or if you'd prefer to fold it in here. Nice work on the diff cleanup! @pythoncheng for visibility.

qq1162210866:
> Thanks @qq1162210866 for the detailed follow-up — the en/zh.json re-serialization was a great catch, and getting the diff down from +2709/−1988 to +778/−57 makes this much easier to review.
> 
> On the three points:
> 
> 1. **PR size** — I'm convinced. The remaining slice is genuinely a vertical dependency chain (Python skip → Go API → frontend button), and splitting it would create dead code in intermediate PRs as you said. Keeping it as one PR is the pragmatic choice. ✅
> 2. **Checkpoint format hardening** — I'd prefer this as a small **follow-up PR** rather than expanding this one further. The defensive warn→delete→miss behavior already prevents crashes, and adding a `version` field + mismatch-as-miss is cheap, isolated hardening that keeps this PR focused. If you open it as a follow-up, I'll review promptly. (Happy either way if you'd rather include it here.)
> 3. **Environment drift on resume** — agreed that stamping tool/config/target hashes and invalidating on mismatch is the right shape, and also better as a follow-up. For now documenting "start a new task when the environment changed" as the safe path is reasonable.
> 
> So: this PR looks good to me as-is. My only ask before merge is the checkpoint `version`-field hardening as a follow-up (not blocking). Ping me when the follow-up is up, or if you'd prefer to fold it in here. Nice work on the diff cleanup! @pythoncheng for visibility.

Folded the checkpoint version-field hardening in here as you offered (01a548cb), since the format isn't merged yet — this way it ships correct from day one and we don't touch the same files twice:

 - Stage files now carry {"stage_id", "version": 1, "content"}.
 - On load, a version mismatch (including old files without the field) reuses the existing defensive path: warn → delete → treat as miss → re-run the stage. A future schema bump just degrades to re-running stages, same as the corruption
   case.
 - Applied identically in all three modules (skill-scan / mcp-scan / agent-scan), smoke-tested for: normal hit, version=0, missing version field, corrupt JSON, missing file.

 Also merged the latest upstream/main into this branch (10f436c3) to keep it current — mostly to pick up #564, which supersedes our earlier mcp-tools version-probing fix for the same headers issue. Conflict resolution there: adopted
 upstream's streamable_http_client(url, http_client=...) form and dropped our _MCP_MAJOR branching; kept our small addition of closing the caller-provided httpx.AsyncClient in a finally (the SDK doesn't take ownership of a client passed
 via http_client).

 Net diff is now +776/−52 across 28 files (shrunk further since our superseded mcp-tools changes no longer count against upstream). The environment-hash invalidation I'll keep as a separate follow-up as agreed — will ping you when it's
 up.

boy-hack:
Thanks for folding in the checkpoint `version`-field hardening, @qq1162210866 — shipping it correct from day one (before the format is merged) is the right call, and applying it identically across all three modules (skill-scan / mcp-scan / agent-scan) with smoke tests for hit / `version=0` / missing-field / corrupt-JSON / missing-file is exactly the defensive behavior we want. Reusing the existing warn→delete→miss path for a future schema bump is clean.

Also good that you merged upstream/main to pick up #564 and dropped the now-superseded mcp-tools version-probing in favor of the upstream `streamable_http_client(url, http_client=...)` form — and keeping the caller-provided `httpx.AsyncClient` closed in a `finally` is a real correctness improvement the upstream form doesn't give you. Net diff shrinking to +776/−52 is much easier to review.

So: this PR looks good to me. **Ready to merge once CI is green.** Environment-hash invalidation as a separate follow-up is fine — ping me when it's up. Nice work on the cleanup and the hardening! @pythoncheng for visibility.

PR #511, opened 2026-07-30
feat(api-checker): integrate API relay checker

kexinoh:
## Summary

Integrates the API relay checker into AI-Infra-Guard as one feature PR:

- adds the Go proxy and WebSocket fallback integration;
- adds the Python service, detection algorithms, reference data, and algorithm-focused unit tests;
- adds the unified `aig check-api` CLI command;
- embeds the checker into the production Agent container and updates Docker/CI configuration;
- adds user, API, parameter, and troubleshooting documentation.

## Review notes

- The work is kept in one PR but remains organized as seven logical commits: proxy, reference data, credential-label fix, service/algorithms, CLI, Docker/CI, and docs.
- The former large proxy test file is split into configured-routing tests and transport/protocol tests. Algorithm behavior is covered under `services/api_checker/tests/test_algorithms.py`.
- `services/api_checker/Dockerfile` is development-only. Production runs the checker as part of `Dockerfile_Agent`; CI does not publish a separate checker image.

## Verification

- `python3 -m unittest discover -s services/api_checker/tests -p test_*.py`: 102 passed
- `go test ./common/apichecker`: passed
- WebSocket fallback test: passed
- `go test ./cmd/cli/cmd`: passed
- both Compose files render successfully; startup shell and workflow YAML validate
- production Agent image build, startup, health endpoint, and embedded Agent + checker processes were verified on the combined implementation

PR #553, opened 2026-08-18
data(api-checker): add reference distributions and charts

kexinoh:
## Scope

- adds the generated PAMELA reference distribution dataset
- adds the four generated model-distribution charts
- keeps generated data separate from implementation so the algorithm PR remains reviewable
- encodes credential-like JSON words with equivalent Unicode escapes to avoid a CodeCC false positive; decoded JSON is unchanged

## Split series

Part 2 of the API Checker split. Part 1 is #511 (Go proxy). The service/algorithm PR consumes these assets but this path-only PR can merge independently.

## Validation

- parsed `distributions.json` with Python JSON tooling
- asserted decoded JSON equality with the original #511 dataset
- `git diff --check`


kexinoh:
Consolidated back into #511 at the author request. Closing this split PR; its logical commit and changes are preserved in #511 for unified review.

PR #569, opened 2026-08-19
fix(data): prevent recurring case-colliding vulnerability rules

ZedingZhang:
## Summary

- remove 10 duplicate vulnerability-rule paths that differ only by directory casing
- keep the repository's normalized product-directory casing without changing any YAML content
- make `yamlcheck` reject distinct YAML paths that collide on case-insensitive filesystems
- run the new `yamlcheck` unit tests in the existing YAML validation workflow

## Problem

Git can track paths that differ only by case, while common Windows and macOS filesystems cannot represent both paths faithfully. The current tree contains 10 such pairs. Every pair points to identical blobs, but Linux checks out both copies and the vulnerability-rule loader walks and loads every YAML file, so the duplicated paths can cause the same rule to be ingested twice.

The collisions are also awkward for Windows contributors: checkout collapses each pair to a single physical file even though both entries remain in the Git index.

## Cleanup

The data-only commit removes the redundant path variants and keeps the existing normalized directories:

| Removed variant | Kept variant | Rules |
| --- | --- | ---: |
| `data/vuln/Dify` | `data/vuln/dify` | 2 |
| `data/vuln/litellm` | `data/vuln/LiteLLM` | 3 |
| `data/vuln/LobeChat` | `data/vuln/lobechat` | 1 |
| `data/vuln_en/litellm` | `data/vuln_en/LiteLLM` | 3 |
| `data/vuln_en/LobeChat` | `data/vuln_en/lobechat` | 1 |

Before removal, each deleted path was verified to have the same Git blob ID as its retained counterpart.

## Regression prevention

`yamlcheck` now:

- cleans paths, normalizes separators, and groups them using case-folded keys
- reports a collision only when a folded key has more than one distinct original path
- ignores repeated occurrences of the exact same path
- sorts groups and paths for deterministic output
- exits before YAML parsing when a collision is found

The unit tests cover directory-case collisions, three or more variants, deterministic ordering, repeated identical paths, distinct roots, and different filenames. The YAML workflow runs these tests before building the checker.

This follows the direction of earlier repository cleanups such as [#236](https://github.com/Tencent/AI-Infra-Guard/pull/236) and [a7d5388](https://github.com/Tencent/AI-Infra-Guard/commit/a7d5388).

## Verification

- `go test -count=1 ./cmd/yamlcheck` — passed
- portable Go 1.23.2 build of `yamlcheck` — passed
- `yamlcheck data/fingerprints data/vuln data/vuln_en` — 4,174 passed, 0 failed
- case-insensitive grouping of all `git ls-files` paths — 0 collision groups
- `git diff --check main...HEAD` — passed

`go test ./...` still has pre-existing Windows/environment-dependent failures on an otherwise clean upstream baseline: undefined test-only symbols in `common/agent`, working-directory assumptions, a required local HTTP service, Chromium helper antivirus blocking, fixed `/mcp-server` paths, and localized-message expectations. No new failure appears in the changed scope, and `cmd/yamlcheck` passes independently.


boy-hack:
Thanks @ZedingZhang for this thorough cleanup! I verified the head tree: after the removals there are **0 remaining case-collision groups** across all 4,028 vuln files, so no rule data is lost — the kept lowercase/`LiteLLM` variants carry the identical blobs. The fix is correct and safe.

**What looks good:**
- The collision detector is well-designed: it normalizes separators (`filepath.ToSlash` + `Clean`), folds case, ignores exact-duplicate paths, and sorts both groups and variants for deterministic output.
- Exiting *before* YAML parsing on collision is the right call — it fails fast and keeps the signal clear.
- Tests cover the important cases (directory case diff, 3+ variants, deterministic ordering, exact-duplicate ignore, distinct roots, different filenames). Good coverage.
- Wiring `go test ./cmd/yamlcheck` into `yaml-lint.yml` before the build step closes the regression loop.

**Minor suggestions:**
1. The new check lives in `main()` and calls `os.Exit(1)` directly, so it's not exercised by `main_test.go` (which only tests the pure function). That's fine — but if you later want CI to catch a regression in the *exit path* too, a `TestMain` wrapper could assert the exit code. Not required.
2. Consider also running the check against `data/fingerprints` (the PR description mentions it in the manual command, but the workflow only builds/checks the vuln dirs). If `fingerprints` can also develop case variants, adding it to the CI invocation would be consistent. Low priority.
3. `strings.Join(collision, " | ")` for the stderr message is readable; consider also printing the *folded* key so the user immediately sees which normalized name collided. Tiny UX improvement.

None of these block merge. The core change is clean, verified, and well-tested. LGTM once CI is green!

ZedingZhang:
Thanks @boy-hack for the detailed review and verification!

Just to clarify suggestion 2: the workflow already runs `./yamlcheck data/fingerprints data/vuln data/vuln_en`, so the case-collision check covers `data/fingerprints` as well.

I’ll keep the exit-path test and folded-key output as possible follow-ups so this PR remains focused. The YAML Lint run is currently awaiting maintainer approval for this fork PR.

boy-hack:
Thanks for the clarification, @ZedingZhang! Good catch — I missed that the workflow already invokes `./yamlcheck data/fingerprints data/vuln data/vuln_en`, so `data/fingerprints` is already covered by the case-collision check. My suggestion 2 is moot. 👍

Keeping the exit-path test and the folded-key output as focused follow-ups makes sense to avoid scope creep. The core change remains clean, verified (0 collision groups across 4,028 vuln files), and well-tested. **LGTM once CI is green!**

PR #578, opened 2026-08-20
fix(agent-scan): quote YAML frontmatter description to prevent parse error

NY1024:
## Problem

Running Agent Scan with the `agentic-supply-chain-detection` skill produces a YAML frontmatter parse error:

```
Error in user YAML: (<unknown>): mapping values are not allowed in this context at line 2 column 47
```

As a result, the skill file fails to load and the agent cannot obtain the skill content, falling back to a degraded mode.

## Root Cause

The `description` field in `agent-scan/agent_scan/prompt/skills/agentic-supply-chain-detection/SKILL.md` contains a bare colon:

```yaml
description: Detect agentic supply-chain risks: compromised dependencies, malicious plugins/tools/models, and untrusted update sources.
```

YAML parsers interpret the second colon (`risks:`) as a nested mapping indicator, triggering the "mapping values are not allowed in this context" error.

## Fix

Wrap the `description` value in double quotes so the colon is treated as part of the string:

```yaml
description: "Detect agentic supply-chain risks: compromised dependencies, malicious plugins/tools/models, and untrusted update sources."
```

## Verification

Validated all 15 SKILL.md files under `agent-scan/agent_scan/prompt/skills/` with `yaml.safe_load` — all pass after the fix.

## Scope

Single-line change, one file only.

### DIFF
--- a/agent-scan/agent_scan/prompt/skills/agentic-supply-chain-detection/SKILL.md
+++ b/agent-scan/agent_scan/prompt/skills/agentic-supply-chain-detection/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: agentic-supply-chain-detection
-description: Detect agentic supply-chain risks: compromised dependencies, malicious plugins/tools/models, and untrusted update sources.
+description: "Detect agentic supply-chain risks: compromised dependencies, malicious plugins/tools/models, and untrusted update sources."
 allowed-tools: dialogue
 ---
 


boy-hack:
Thanks @NY1024 for the thorough root-cause analysis — the explanation of why the bare colon trips the YAML parser ("mapping values are not allowed") is spot on, and validating all 15 SKILL.md files with `yaml.safe_load` is a nice touch.

One heads-up: this is a **duplicate of PR #580** by @cloudbypriyank — both fix the exact same line in the same file (and #580 explicitly Closes #577). Could you two coordinate and consolidate into a single PR to avoid merge conflicts? Either diff is correct; keeping one is enough. Thanks!

NY1024:
Thanks @boy-hack for the heads-up! Confirmed this is a duplicate of PR #580 by @cloudbypriyank, which also explicitly Closes #577. Closing this one to avoid merge conflicts — #580 is the right one to merge. Thanks @cloudbypriyank for the parallel fix!

PR #588, opened 2026-08-21
fix(parser): keep pre-release versions below their release in versionCheck

pucedoteth:
## What

`versionCheck` (`common/fingerprints/parser/synax.go`) strips every letter run from a version string. When the letters sit **between two digits**, the digits on either side merge:

```
"3.11.0rc2"  ->  "3.11.02"   // = 3.11.2
"0.6.0rc1"   ->  "0.6.01"    // = 0.6.1
```

So a pre-release normalizes to a version *higher* than the release it precedes. Every advisory rule that compares versions is evaluated against that value, which inverts the result for targets running a pre-release build. Measured through `AdvisoryEval` before the fix:

| rule | scanned version | result | |
|---|---|---|---|
| `version < "3.11.0"` | `3.11.0rc2` | `false` | advisory missed |
| `version <= "3.11.1"` | `3.11.0rc2` | `false` | advisory missed |
| `version == "3.11.2"` | `3.11.0rc2` | `true` | advisory reported for a version the target isn't running |

## Why

`X.Y.ZrcN` / `X.Y.ZbN` / `X.Y.ZaN` is the standard PEP 440 pre-release form, and the AI infrastructure this scanner targets is largely Python (vLLM, Ray, Gradio, …), where release candidates are commonly deployed. 1920 rule files under `data/vuln` and `data/vuln_en` reference version comparisons, so the miss applies across the whole advisory set rather than to one rule.

## How

Rewrite a letter run between two digits as a pre-release separator before the letters are stripped, so `3.11.0rc2` becomes `3.11.0-2` — which `hashicorp/go-version` orders below `3.11.0`, as it should be.

Versions whose digits are not separated by letters are untouched; the added table pins the existing shapes (`1.2.3`, `v1.2.3`, `1.2.3b`, `1.2.3-rc1`, `1.2.3.RELEASE`, `latest`, `""`) so the normalization stays put.

Both new tests fail on `main`:

```
versionCheck("3.11.0rc2") = "3.11.02", want "3.11.0-2"
rule version < "3.11.0" with version "3.11.0rc2" = false, want true
```

`go test ./common/fingerprints/parser/` passes with the change; `gofmt` and `go vet` are clean.

Out of scope, mentioned for the record: the dotted Java form `1.0.0.RC1` still normalizes to `1.0.0.01` through the separate `\.[A-Za-z]+` → `.0` rule. Fixing that means changing how `.RELEASE`-style suffixes are handled, which is a wider behaviour change — happy to follow up if you want it.

Note on the test run: `internal/mcp/utils`, `common/websocket`, `common/fingerprints/preload` and `common/utils` already fail on a clean checkout in a local environment (they need `/mcp-server` paths, a listener on `127.0.0.1:8265`, cwd-relative `data/` files, and Chinese-locale API strings). I verified those failures are present without this change.

---

*Disclosure: written with AI assistance; the analysis, fix and tests were verified locally.*

### DIFF
--- a/common/fingerprints/parser/synax.go
+++ b/common/fingerprints/parser/synax.go
@@ -282,6 +282,9 @@ func (r *Rule) Eval(config *Config) bool {
 	return evalExpr(r.root, config)
 }
 
+// preReleaseSuffixRe 匹配夹在两个数字之间的字母段，例如 "3.11.0rc2" 中的 "rc"。
+var preReleaseSuffixRe = regexp.MustCompile(`([0-9])[A-Za-z]+([0-9])`)
+
 // versionCheck 版本号格式标准化处理
 // 输入版本号字符串，返回处理后的版本号字符串
 // 去除版本号中的字母并进行格式统一化
@@ -293,7 +296,10 @@ func versionCheck(version string) string {
 	// 正则替换所有单词
 	compile := regexp.MustCompile(`[A-Za-z]+`)
 	if compile.MatchString(version) {
-		newVersion := regexp.MustCompile(`\.[A-Za-z]+`).ReplaceAllString(version, ".0")
+		// 字母位于两个数字之间时（如 PEP 440 的 "3.11.0rc2"）先改写为预发布号，
+		// 否则直接删除字母会把两侧数字粘连成 "3.11.02"，即 3.11.2。
+		newVersion := preReleaseSuffixRe.ReplaceAllString(version, "$1-$2")
+		newVersion = regexp.MustCompile(`\.[A-Za-z]+`).ReplaceAllString(newVersion, ".0")
 		newVersion = compile.ReplaceAllString(newVersion, "")
 		//gologger.Debugf("version:%s=>%s", version, newVersion)
 		version = newVersion
--- a/common/fingerprints/parser/synax_test.go
+++ b/common/fingerprints/parser/synax_test.go
@@ -150,3 +150,61 @@ func TestEval(t *testing.T) {
 		}
 	}
 }
+
+// TestVersionCheckPreRelease 覆盖 PEP 440 风格的预发布版本号，
+// 例如 "3.11.0rc2"：直接删除字母会把两侧数字粘连成 "3.11.02"（即 3.11.2）。
+func TestVersionCheckPreRelease(t *testing.T) {
+	for _, c := range []struct {
+		version string
+		want    string
+	}{
+		{"3.11.0rc2", "3.11.0-2"},
+		{"0.6.0rc1", "0.6.0-1"},
+		{"1.2.3b1", "1.2.3-1"},
+		// 未被字母分隔的数字保持原样
+		{"1.2.3", "1.2.3"},
+		{"v1.2.3", "1.2.3"},
+		{"1.2.3b", "1.2.3"},
+		{"1.2.3-rc1", "1.2.3-1"},
+		{"1.2.3.RELEASE", "1.2.3.0"},
+		{"latest", "999"},
+		{"", "0"},
+	} {
+		if got := versionCheck(c.version); got != c.want {
+			t.Fatalf("versionCheck(%q) = %q, want %q", c.version, got, c.want)
+		}
+	}
+}
+
+// TestAdvisoryEvalPreRelease 预发布版本必须落在其正式版本之前，
+// 否则受影响的目标会被漏报。
+func TestAdvisoryEvalPreRelease(t *testing.T) {
+	for _, c := range []struct {
+		rule    string
+		version string
+		want    bool
+	}{
+		{`version < "3.11.0"`, "3.11.0rc2", true},
+		{`version <= "3.11.1"`, "3.11.0rc2", true},
+		{`version == "3.11.2"`, "3.11.0rc2", false},
+		{`version < "0.6.0"`, "0.6.0rc1", true},
+		// 正式版本不受影响
+		{`version < "3.11.0"`, "3.10.9", true},
+		{`version < "3.11.0"`, "3.11.0", false},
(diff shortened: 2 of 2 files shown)


boy-hack:
Thanks @pucedoteth for the clear root-cause analysis — `3.11.0rc2` normalizing to `3.11.02` (= 3.11.2) is a nasty silent inversion, and with ~1920 version-comparison rules it affects the whole advisory set, not one rule. The fix (rewrite the letter run between two digits as a pre-release separator `-` before stripping) is correct and `hashicorp/go-version` orders `3.11.0-2` below `3.11.0` as expected.

The test table is comprehensive (`3.11.0rc2`, `0.6.0rc1`, `1.2.3b1`, `1.2.3-rc1`, `1.2.3.RELEASE`, plus the `AdvisoryEval` miss/hit cases) and covers both the normalization and the end-to-end eval.

One edge case to be aware of (not a blocker): a pre-release with no trailing digit, e.g. `3.11.0rc`, won't match `([0-9])[A-Za-z]+([0-9])` and just strips to `3.11.0` — i.e. treated as the release rather than below it. Rare in practice (PEP 440 pre-releases almost always carry a number), but worth a note. Also `latest`→`999` and `""`→`0` behavior is unchanged, good.

Looks good to merge.

pucedoteth:
Thanks for the review — and confirmed on `3.11.0rc`. `([0-9])[A-Za-z]+([0-9])` requires the trailing digit, so a bare-suffix pre-release never reaches the rewrite, falls through to the letter strip, and compares equal to `3.11.0` instead of below it.

Worth flagging that it isn't a one-character follow-up. Relaxing the tail to `([0-9]*)` makes the `$1-$2` replacement emit a dangling `3.11.0-`, so handling it properly means carrying the label through (`3.11.0-rc`) rather than widening the match — a bigger change than the digit-splicing inversion this PR is scoped to, and one that wants its own test table.

Happy to fold it in here if you'd rather have both in one change; otherwise I'll send it as a follow-up once this lands.

boy-hack:
Thanks @pucedoteth — confirmed, and your reasoning is sound. The bare-suffix case (`3.11.0rc` → dangling `3.11.0-`) is genuinely a separate change: it needs the label carried through (`3.11.0-rc`) rather than just widening the regex, plus its own test table. Folding it here would blur the scope of an otherwise tight, well-tested fix.

I'd prefer the follow-up PR — keep this one scoped to the digit-splicing inversion (the actual bug affecting ~1920 rules), and land the pre-release-label handling separately so each change stays independently reviewable and bisectable.

This PR looks good to merge as-is. 🚀

pucedoteth:
Agreed on keeping the scope split — thanks for the quick turnaround on both rounds.

I'll send the bare-suffix follow-up (`3.11.0rc` -> `3.11.0-rc`) as its own PR once this one lands, so it rebases cleanly on the `$1-$2` rewrite rather than stacking on an unmerged branch. It'll carry the label through instead of widening the match, with its own table covering bare suffixes (`rc`, `alpha`, `beta`), the existing digit-suffixed cases as regression guards, and the ordering assertions against `3.11.0` and `3.11.1`.

Nothing needed on this PR from my side.

boy-hack:
Thanks @pucedoteth — confirmed and agreed on the scope split. Your analysis that `3.11.0rc` (bare suffix, no trailing digit) is a genuinely separate change is correct: it needs the pre-release label carried through (`3.11.0-rc`) rather than just widening the `([0-9])[A-Za-z]+([0-9])` match (which would emit a dangling `3.11.0-`), and it deserves its own test table. Folding it into this PR would blur the scope of an otherwise tight, well-tested fix for the digit-splicing inversion.

This PR is good to merge as-is. 🚀 When you send the bare-suffix follow-up, please:
- link it back to this PR and to the issue it addresses, and
- include the bare-suffix cases (`rc`, `alpha`, `beta`) plus the existing digit-suffixed cases as regression guards, with ordering assertions against `3.11.0` / `3.11.1` (as you outlined).

pucedoteth:
Thanks @boy-hack — follow-up is up as #626.

It does what you asked for: the label is carried through (`3.11.0rc` -> `3.11.0-rc`) rather than the match being widened, so there's no dangling `3.11.0-`. It links back here, and the test table keeps the existing digit-suffixed cases as regression guards alongside the new bare-suffix ones (`rc`, `alpha`, `beta`), with ordering assertions against `3.11.0` and `3.11.1`.

Two things there worth your attention, both called out in the PR body:

1. **It's stacked on this branch**, so it currently shows both commits. It reduces to one once this merges — please merge this one first.

2. **It changes output for cases this PR already handles**: `3.11.0rc2` becomes `3.11.0-rc2` rather than `3.11.0-2`. Ordering against the release is identical; the label is just kept instead of discarded. The reason for going further than the bare-suffix case is that keeping only the number cannot order labels against each other — `3.11.0a1` and `3.11.0rc1` both normalize to `3.11.0-1` today and compare *equal*. Keeping the label fixes that too. If you'd rather I hold the existing outputs fixed and touch only the bare-suffix path, say so and I'll narrow it.

I also found that `1.2.3-rc` currently normalizes to a degenerate `1.2.3-`; the same change fixes it.

boy-hack:
@pucedoteth Thanks — confirmed #626 is up and links back here as the follow-up, with the bare-suffix label carried through (`3.11.0rc` -> `3.11.0-rc`) rather than widening the match. Good.

This PR (#588) remains good to merge as-is. One merge-ordering note for the maintainer: please merge **#588 first**, then rebase **#626** onto `main` — #626 is currently based on `main` and includes this commit, so once #588 lands it should reduce to its single follow-up commit (the regex is redefined, not duplicated, so there's no compile conflict, but the extra commit should be dropped).

Issue #572, opened 2026-08-19
扫描包含非 UTF-8 文件名的仓库时出现 UnicodeEncodeError


Yattoo:
扫描某些仓库时，若仓库中存在文件名包含非 UTF-8 字节、编码异常或损坏的文件名，扫描过程会失败并退出。

报错信息：

[...]

PR #589, opened 2026-08-21
fix(mcp-scan): strip lone surrogates from non-UTF-8 filenames

rootkiller6788:
## Problem

Scanning a project that contains a non-UTF-8 filename aborts mcp-scan with `UnicodeEncodeError` / `PydanticSerializationError`. Python's `os.walk()`/`os.listdir()` decode such filenames into lone surrogate characters (`\udc00`-`\udfff`, PEP 383 `surrogateescape`), and pydantic's `model_dump_json()` cannot serialize them.

This is the same bug class reported in #572 for skill-scan (fixed in PR #571). mcp-scan has the same failure modes and is still vulnerable:
- `agent.py:_build_repo_tree` — `os.walk()` on the scanned repo feeds filenames into the repo tree shown to the LLM
- `utils/pre_scan.py` — `os.walk()` filenames end up in the security-hint text
- `utils/aig_logger.py:_log` — `AgentMsg(...).model_dump_json()` raises `PydanticSerializationError` on any surrogate reaching the structured logger
- `tools/dispatcher.py:_format_result` — tool results with surrogate filenames crash later serialization

Triggering path: `McpTask`/`AgentTask` spawn `python mcp-scan/main.py --aig-mode` (or an MCP scan against a repo containing such a file). One `\xff` byte in a filename is enough to kill the whole scan.

## Fix

Add a recursive `strip_surrogates()` helper in `mcp_scan/utils/__init__.py` (replaces lone surrogates with `?` via `encode('utf-8', 'replace')`), and apply it:
1. At the source — `_build_repo_tree` and `pre_scan` outputs
2. Defensively in the AIG logger sink (`_log`) before `model_dump_json()`
3. In the tool-result formatter

Same approach as the accepted skill-scan fix (PR #571).

## Tests

`pytests/test_surrogate_sanitization.py` covers the helper, the logger sink, and (POSIX-only, skipped on Windows) real non-UTF-8 files through `pre_scan` and `_build_repo_tree`. Full suite: 13 passed, 2 skipped.

### DIFF
--- a/mcp-scan/mcp_scan/agent/agent.py
+++ b/mcp-scan/mcp_scan/agent/agent.py
@@ -22,6 +22,7 @@ from typing import Any
 
 from mcp_scan.agent.base_agent import BaseAgent
 from mcp_scan.tools.dispatcher import ToolDispatcher
+from mcp_scan.utils import strip_surrogates
 from mcp_scan.utils.aig_logger import mcpLogger
 from mcp_scan.utils.extract_vuln import VulnerabilityExtractor, extract_result
 from mcp_scan.utils.loging import logger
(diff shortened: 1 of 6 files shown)


boy-hack:
Thanks @rootkiller6788 for porting the surrogate-stripping fix from skill-scan #571 to mcp-scan — this is the same `UnicodeEncodeError`/`PydanticSerializationError` crash class and mcp-scan definitely had the same exposure points (`_build_repo_tree`, `pre_scan`, the AIG logger sink, `_format_result`).

The recursive `strip_surrogates()` helper and applying it both at the source (repo tree / pre-scan hint) *and* defensively in the logger sink is the right layered approach — a single missed call site won't kill the scan. Tests are thorough, including the POSIX raw-byte filename cases.

Minor notes (non-blocking):
1. `strip_surrogates` recurses into dict **values** but not **keys**. If a surrogate ever lands in a dict key it'd be skipped — unlikely for filenames, but worth a one-line `k` handling if you want full coverage.
2. `mcp_scan/utils/__init__.py` already existed (license header only); adding the helper there is fine and keeps the import path clean.

Looks good to merge once CI is green.

PR #599, opened 2026-08-25
fix(docker): use Tencent Debian mirror for agent build

kexinoh:


boy-hack:
Thanks @kexinoh. Minimal, targeted fix: declare `ARG DEBIAN_MIRROR` (default `https://mirrors.cloud.tencent.com`) in the python stage and `sed` the deb822 source in `Dockerfile_Agent`.

**Verification:**
1. **Correct target file:** `python:3.12-slim` (Debian 12 Bookworm) uses `/etc/apt/sources.list.d/debian.sources` with `http://deb.debian.org` URIs, so the `sed` matches. Good.
2. **Scope is right:** `ARG DEBIAN_MIRROR` is declared *after* `FROM python:3.12-slim`, so it's in scope for the `RUN sed` in the same stage — `--build-arg DEBIAN_MIRROR=...` override will work. (An ARG before the first FROM wouldn't carry into the stage, but that's not the case here.)
3. **`set -eux` safe:** `sed` exits 0 even with no match, so the build won't break on images where the URI differs.
4. **Resulting URI:** `http://deb.debian.org/debian` → `https://mirrors.cloud.tencent.com/debian`, which is the correct Tencent Debian mirror path.

This should speed up the agent image build in Tencent CI. LGTM — no behavioral change to the scan logic.

PR #601, opened 2026-08-25
Fix/report language localization

Aj7Ay:
Reports were not consistently rendered in the language selected in the UI. This PR fixes the two biggest gaps and one smaller one:

1. AI Infra Scan - CVE description/recommendation/severity text was always Chinese, regardless of the selected language. Root cause: the agent-side scan always loads vulnerability data via LoadFromHost → the webserver's /api/v1/knowledge/vulnerabilities endpoint, which hardcoded data/vuln (Chinese) with no language parameter at all - request.Language never reached this path. Now the language flows through end-to-end and data/vuln_en is served when English is selected.
2. Jailbreak Evaluation - Original Input, Attack Input, and Model Output rendered in whatever language the dataset/target model happened to use, regardless of the selected report language. Added a generic, bidirectional translation step (_translate_text/_a_translate_text) that:
   - translates into whichever language is selected (not just Chinese, which was the only direction the old code supported),
   - skips text already in the target language,
   - skips text whose language can't be confidently judged (e.g. Zalgo-obfuscated attack payloads), so obfuscated/non-linguistic content is never mangled,
   - is cached per scan run, since the same dataset scenario text is frequently reused across many attack techniques/vulnerability types,
   - does not re-translate reason, which is already correctly localized earlier in the pipeline.
3. Agent Scan - task-list step titles ("Info Collection", "Vulnerability Detection", "Vulnerability Review") were hardcoded English even in Chinese mode, unlike MCP Scan which already localized these. Now matches MCP Scan's pattern.

Not in scope (called out, not silently skipped)

- The Knowledge Base "create/edit vulnerability" endpoints still only write data/vuln (Chinese) - an admin editing or adding a CVE entry won't get an English version automatically. Left as a follow-up.
- The OWASP ASI category label in Agent Scan (e.g. "Agent Goal Hijack") stays English-only in all languages - treated as a fixed taxonomy term, same as "CVE"/"CVSS" elsewhere, not report prose.
- Jailbreak translation is not parallelized - it's cached, but still sequential per test case, so large runs (hundreds of cases) will see added latency from the extra translation calls.

### DIFF
--- a/AIG-PromptSecurity/deepteam/red_teamer/red_teamer.py
+++ b/AIG-PromptSecurity/deepteam/red_teamer/red_teamer.py
@@ -115,8 +115,6 @@ class RedTeamer:
         self.simulator_model, _ = initialize_model(simulator_model)
         self.evaluation_model, _ = initialize_model(evaluation_model)
         self.async_mode = async_mode
-        # 缓存已翻译过的文本，避免同一条数据集原文/相同 reason 在多个测试用例间重复翻译
-        # （同一批次的越狱数据集通常会被多种攻击手法/多个漏洞类型重复引用同一条 original_input）
         self._translation_cache: Dict[str, str] = {}
         self.synthetic_goldens: List[Golden] = []
         self.custom_metric = None  # 添加自定义metric属性
@@ -151,7 +149,6 @@ Direct translation without separators"""
         return f"Translate to {logger.lang} (output translation only):\n\n{text}"
 
     def _lang_category(self) -> str:
-        """把 logger.lang（如 'en'/'zh'/'zh_CN'）映射到 judge_language() 的词表上"""
         lang = (logger.lang or "").lower()
         if lang.startswith("zh"):
             return "chinese"
@@ -160,9 +157,6 @@ Direct translation without separators"""
         return "default"
 
     def _translate_text(self, text: str) -> str:
-        """把任意报告文本翻译成当前配置的报告语言（同步版本）。
-        已经是目标语言的文本、以及无法判断语言的文本（如混淆/编码后的攻击载荷）都会原样跳过，
-        避免把非自然语言内容（比如 Zalgo 编码）误翻译损坏。"""
         if not isinstance(text, str) or not text.strip():
             return text
         target = self._lang_category()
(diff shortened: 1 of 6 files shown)


boy-hack:
Thanks @Aj7Ay for this comprehensive localization pass. The combination of the Go-side `taskTitles` localization plus threading `Language` through `executeScan` → `LoadFromHost` (en → `data/vuln_en`) and `knowledge_api.go` (en → `data/vuln_en`), together with the Python-side `_translate_text` (language detection + per-instance cache) in `red_teamer`, is a coherent design. A few notes:

**Good:**
- `_translation_cache` keyed by `f"{target}:{text}"` with the `judge_language` skip for already-target / unjudgeable text (Zalgo/encoded payloads) is exactly right — it avoids corrupting attack payloads during translation.
- `originalInput` for `RedTeam` cases is `None` → `_translate_text(None)` returns `None` (not a str) with no crash. Safe.

**Points to confirm:**
1. **Double-translation risk:** `get_risk_assessment_json` now translates `input`/`output`/`reason` directly. Please grep to confirm `case.reason` isn't *also* run through the old `_translate_reason` path elsewhere — otherwise the reason could be translated twice.
2. **Unbounded cache:** the cache is per-`RedTeamer` instance (per scan), so memory is bounded by one scan's dataset — acceptable, just noting it's instance-scoped, not global.
3. **Missing PR description:** the body is empty. Please add a short summary + how to test (e.g. run a red-team scan with `logger.lang="en"` and confirm report fields + task titles are English). Helps reviewers.

Overall solid. Ship it after adding the description — and please double-check point 1.

Aj7Ay:
@boy-hack Updated, please check.

1. **Double-translation risk** - confirmed and fixed in 83c91cd4. `case.reason` was going through `_translate_reason`/`_a_translate_reason` at measurement time, then again through `_translate_text` in `get_risk_assessment_json`. Removed the second call - `reason` is already correctly localized by the time it reaches the choke point (either via the measurement-time translation, or via `logger.translated_msg`/gettext for the canned error-path reasons), so no correctness issue in practice, but it was redundant and could trigger an unnecessary extra LLM call if `judge_language` misread the already-translated text. `originalInput`/`input`/`output` still go through `_translate_text` since nothing translates those earlier.
2. Cache scope - agreed, left as instance-scoped, no change needed.

PR #603, opened 2026-08-26
feat(api-checker): add Gemini, Gemma, and GLM fingerprints

kexinoh:
## Summary

- add 3 Gemini, 7 Gemma, and GLM-5.3 model fingerprints
- include 500 valid samples for each new baseline
- update bundled-baseline and detectable-model contract tests from 29 to 40 models

## Validation

- `python3 -m pytest services/api_checker/tests -q` — 102 passed
- `python3 -m services.api_checker.main list` — lists all 40 models
- baseline integrity, uniqueness, diff check, and secret scan passed


boy-hack:
Thanks @kexinoh — solid addition. 11 new model baselines (29 → 40), each with 500 samples, and both the bundled-baseline and detectable-model contract tests are bumped and subset-asserted. A few notes:

1. **Model-name reality check (minor):** `google/gemma-4-26b-a4b-it`, `google/gemma-4-31b-it`, and `google/gemini-3.1-flash-lite` are forward-looking names. I'm assuming they're real as of the 2026-08-25 baseline timestamp, but please confirm the fingerprint samples were actually collected from live endpoints and not placeholders — a wrong model id silently breaks detection for that entry.

2. **Baseline integrity is the real guardrail:** since `baselines.json` is pure data, the key thing is that the uniqueness / diff / integrity checks in CI stay green. Your validation says they pass — good. Consider tightening `test_bundled_baselines_are_complete` to also assert `len(baseline["rawData"]) == 500` (or `iterations == 500`) so a truncated baseline can't slip in later.

3. **`apiType: openai` for all Gemini/Gemma/GLM entries** — fine if they're hit through an OpenAI-compatible gateway; just flagging that `apiType` drives the request path, so double-check it matches how these are actually served.

Otherwise LGTM. Nice to see the contract tests extended rather than just the data file.

PR #611, opened 2026-08-27
Revert "docs(api-checker): generalize relay consistency check"

kexinoh:
Reverts Tencent/AI-Infra-Guard#607
避免github action不通过

### DIFF
--- a/services/api_checker/README.md
+++ b/services/api_checker/README.md
@@ -7,8 +7,40 @@ AI 模型指纹识别 + 中转站检测工具
 ### 1. 随机数指纹识别（统计方法）
 通过让 AI 模型"随机选数字 1-355"，收集大量样本的分布指纹来区分不同模型。
 
-### 2. 中转站一致性检测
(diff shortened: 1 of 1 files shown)


boy-hack:
@kexinoh — reverting #607 to get CI green is a reasonable short-term unblock, but I'd push back gently: this drops a genuinely useful chunk of documentation (the thinking-signature AEAD analysis, the 11-item detection table). Reverting docs to satisfy a failing action is a band-aid.

Before merging, can you link **which** GitHub Action is failing and why? If it's a test that asserts README content/structure matches the code (e.g. a contract test referencing "中转站一致性检测"), the right fix is to update that assertion, not delete the docs. If it's a docs lint, we should exclude or fix the lint.

If the failure is truly unavoidable right now and this is just to unblock the branch, that's fine — but please open a follow-up to restore the #607 content properly once the root cause is addressed, rather than letting the better docs stay reverted.

(Note: this is independent of #603, so no conflict there.)

PR #615, opened 2026-08-27
docs: explain Tencent Debian mirror build acceleration

kexinoh:
## Summary

- document the Tencent Cloud Debian mirror as an optional Agent image build acceleration
- place the guidance next to the source-build workflow in both Chinese and English
- clarify that Debian suites, components, and `Signed-By` verification remain unchanged
- include commands to rebuild the Agent and restore the official Debian source

## Validation

- Markdown code fences checked
- `git diff --check` passed
- secret scan passed

### DIFF
--- a/aigdocs/docs/getting-started.md
+++ b/aigdocs/docs/getting-started.md
@@ -31,6 +31,23 @@ cd AI-Infra-Guard
 docker-compose up -d
 ```
 
+**可选：使用腾讯云 Debian 软件源加速 Agent 镜像构建**
+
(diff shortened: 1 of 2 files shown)


boy-hack:
Thanks @kexinoh — clear and accurate documentation of an optional build-time accelerator. The guidance correctly preserves suites/components/`Signed-By` and gives a restore command, which is exactly what a maintainer needs.

One small caveat worth adding: `sed` only rewrites `deb.debian.org`, so `security.debian.org` is left on the official CDN. Tencent Cloud's mirror does mirror `deb.debian.org` but coverage of `deb.debian-security` can differ — you may want to mention that security updates stay on the official source, or add a second `sed` for `security.debian.org` if the mirror carries it. Not a doc bug, just a completeness note.

Validation checks (fences, `git diff --check`, secret scan) all good. LGTM for docs.

PR #619, opened 2026-08-28
feat(agent-scan): add OrcaRouter as a first-class provider

lovejones2914-spec:
Adds [OrcaRouter](https://www.orcarouter.ai) as a first-class provider in `agent-scan`, mirroring the existing `openrouter` entry in the provider registries. Users of A.I.G's agent security scanner can now select OrcaRouter by name and get a working endpoint out of the box, instead of treating it as an anonymous custom base URL.

OrcaRouter is an OpenAI-compatible AI gateway built for both models and agents. Like OpenRouter, it exposes a provider/model namespace across many models — but it also combines adaptive routing, automatic failover, zero-markup inference, observability, guardrails, and agent-tool governance behind the same endpoint. Adding `orcarouter` as a first-class provider means agent-scan users can use that stack directly.

It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.

Changes:
- `agent-scan/providers.yaml` and `agent-scan/agent_scan/core/agent_adapter/providers.yaml`: new `orcarouter` provider under `openai_compatible`, with `ORCAROUTER_API_KEY`, `https://api.orcarouter.ai/v1`, and `orcarouter/fusion*` default models.
- `agent-scan/agent_scan/core/agent_adapter/adapter.py`: provider docstring updated.
- `agent-scan/README.md` / `README_zh.md`: provider type list updated.

Validation:
- `pytest pytests/` — 2 passed.
- Live-tested the new provider path through `AIProviderClient` with a real key: `success: True, Status: 200`, response `1`.

Discord: discord.gg/YEubt8enRA · X: https://x.com/OrcaRouter

I'm an engineer on the OrcaRouter team.

### DIFF
--- a/agent-scan/README.md
+++ b/agent-scan/README.md
@@ -127,7 +127,7 @@ targets:
       transform_response: "reply"
 ```
 
-Supported provider types: `http`, `dify`, `coze`, `openai`, `anthropic`, `google`, `cohere`, `huggingface`, `replicate`, `ollama`, `localai`, `litellm`, `websocket`, and more.
+Supported provider types: `http`, `dify`, `coze`, `openai`, `anthropic`, `google`, `cohere`, `huggingface`, `replicate`, `ollama`, `localai`, `litellm`, `openrouter`, `orcarouter`, `websocket`, and more.
 
 #### ⚠️ `transform_response` — Critical for Custom HTTP Providers
 
--- a/agent-scan/README_zh.md
+++ b/agent-scan/README_zh.md
@@ -127,7 +127,7 @@ targets:
       transform_response: "reply"
 ```
 
-支持的 provider 类型：`http`、`dify`、`coze`、`openai`、`anthropic`、`google`、`cohere`、`huggingface`、`replicate`、`ollama`、`localai`、`litellm`、`websocket` 等。
+支持的 provider 类型：`http`、`dify`、`coze`、`openai`、`anthropic`、`google`、`cohere`、`huggingface`、`replicate`、`ollama`、`localai`、`litellm`、`openrouter`、`orcarouter`、`websocket` 等。
 
 #### ⚠️ `transform_response` — 自定义 HTTP Provider 的关键配置
 
--- a/agent-scan/agent_scan/core/agent_adapter/adapter.py
+++ b/agent-scan/agent_scan/core/agent_adapter/adapter.py
@@ -208,7 +208,7 @@ class AIProviderClient:
     - Mistral AI, Groq, Ollama
     - AWS Bedrock
     - Cohere, Deepseek, Perplexity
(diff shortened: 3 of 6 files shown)


boy-hack:
Thanks @lovejones2914-spec for the contribution. Adding `orcarouter` as a first-class `openai_compatible` provider mirrors the existing `openrouter` entry cleanly — updated both `providers.yaml` copies (`agent-scan/providers.yaml` and `agent-scan/agent_scan/core/agent_adapter/providers.yaml`), the adapter docstring, and both READMEs (EN + ZH). Good to keep the two provider registries in sync since they're duplicated.

A few notes, none blocking:
- `env_keys: ["ORCAROUTER_API_KEY"]` — consistent with the other providers' single-key style. Fine.
- `default_model: "orcarouter/fusion"` plus `fusion-flash` / `fusion-mini` — please double-check the exact model namespace against OrcaRouter's current API, since a wrong default model string surfaces as a confusing 404 at scan time rather than a config error.
- The `pytest pytests/` pass (2 tests) only exercises existing paths, not the new provider selection — a tiny smoke test that `orcarouter` resolves to the right `base_url`/`endpoint` would harden this, but the live 200 you reported covers the runtime path.

Looks good to merge. Thanks for the clear validation notes.

lovejones2914-spec:
Thanks @boy-hack for the review. Addressed both notes:

1. **Model namespace verified** — `https://api.orcarouter.ai/v1/models` (no auth required) returns `orcarouter/fusion` (1M ctx), `orcarouter/fusion-flash` (200K ctx), and `orcarouter/fusion-mini` (1M ctx), so the default model string and the `models` list match the live API.
2. **Smoke test added** — new `agent-scan/pytests/test_orcarouter_provider.py` asserts the registry entry resolves to `base_url: https://api.orcarouter.ai/v1`, `endpoint: /chat/completions`, `default_model: orcarouter/fusion`, and the expected `env_keys`/`models`. Full `pytest pytests/` now passes (3 tests, including this one).

boy-hack:
Thanks @lovejones2914-spec for the quick turnaround on both notes.

- ✅ **Model namespace verified** — pulling `https://api.orcarouter.ai/v1/models` to confirm `orcarouter/fusion`, `fusion-flash`, and `fusion-mini` match the registry is exactly the check I wanted. That removes the 404-at-scan-time risk.
- ✅ **Smoke test added** — `test_orcarouter_provider.py` asserting `base_url`/`endpoint`/`default_model`/`env_keys`/`models` is a good guard against a silent registry typo. And you kept both `providers.yaml` copies in sync, which is the easy-to-miss part.

Both points addressed and the new test is in the suite. LGTM — looks good to merge. Thanks for the clean contribution.

PR #623, opened 2026-08-31
fix(mcp-scan): 空输出的扫描被评分为满分形成假阴性，增加完整性标记

valleyxht:
在测试扫描 MCP Server 项目时发现：当项目中存在超大源文件（单个文件 100KB+）时，扫描可能以成功状态结束，但结果是 0 个发现 + 满分 100，而实际上审计过程在中途已经失效。这构成一个假阴性盲区：恶意 MCP Server可以故意在项目里塞入超大文件，干扰审计 Agent 的工作上下文，从而变相绕过代码审计。
检查发现大项目会让审计 Agent 的上下文增长到触发历史压缩。压缩后 Agent 可能丢失已积累的分析状态，以空输出结束。空输出流入 calc_mcp_score([])，该函数对空列表无条件返回 100
修复：在两个结果汇总点增加检查：当审计原始输出与提取的漏洞列表均为空时，将扫描标记为不完整：
  - 记录 warning 日志（扫描可能不完整）
  - 结果元数据设置 `scanNote: "possibly-incomplete"`
  - SARIF 输出在 `properties` 中携带 scanNote

  触发问题的历史压缩目前不保留已发现的安全发现，还可以在 prompt 层面进一步缓解：例如在压缩模板中增加专门的安全发现段落，让总结模型在压缩时携带已确认/疑似的发现。

### DIFF
--- a/mcp-scan/mcp_scan/agent/agent.py
+++ b/mcp-scan/mcp_scan/agent/agent.py
@@ -307,6 +307,19 @@ class ScanPipeline:
         return result
 
 
+def _warn_if_scan_incomplete(raw_output: str, vuln_results: list, result_meta: dict) -> None:
+    """审计原始输出与漏洞列表均为空时，标记扫描可能不完整。
+
+    典型场景：上下文压缩丢失已积累的分析状态后，Agent 直接以空输出结束，
+    空结果会被 calc_mcp_score 评为 100 分，形成"未发现漏洞"的假阴性结论。
+    此处显式打标记，避免空报告被误读为安全。
+    """
+    if vuln_results or (raw_output or "").strip():
+        return
+    logger.warning("扫描可能不完整")
+    result_meta["scanNote"] = "possibly-incomplete"
+
+
 class Agent:
     """aig-mcp-scan's main Agent, entry point for the scan pipeline.
 
@@ -452,6 +465,7 @@ markdown格式返回
                 "results": vuln_results,
             }
         )
+        _warn_if_scan_incomplete(audit_result, vuln_results, result_meta)
         mcpLogger.result_update(result_meta)
         return result_meta
 
@@ -575,6 +589,7 @@ markdown格式返回
                 "results": vuln_results,
             }
         )
+        _warn_if_scan_incomplete(vuln_review, vuln_results, result_meta)
         mcpLogger.result_update(result_meta)
         return result_meta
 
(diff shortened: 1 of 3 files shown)


boy-hack:
Thanks @valleyxht — this is a sharp catch. A "successful" scan that returns 0 findings + score 100 because the agent silently aborted on large files is a genuinely dangerous false-negative, and the `calc_mcp_score([]) -> 100` behavior makes it look clean. Marking `scanNote: "possibly-incomplete"` in `result_meta` and surfacing it in the SARIF `properties` is a good, low-risk mitigation. The pytest coverage (`_warn_if_scan_incomplete`) for empty/whitespace/None output and the "don't mark when findings exist" cases is solid.

Suggestions:
- **Promote the default to `"complete"` consistently**: `sarif_formatter` uses `result_meta.get("scanNote", "complete")` while the agent sets `"possibly-incomplete"`. That's fine, but consider also writing `"scanNote": "complete"` explicitly when the scan *is* complete, so consumers can rely on the key always being present rather than inferring absence == complete.
- **Root-cause follow-up (you already flagged it)**: The trigger is context compaction dropping accumulated findings. The note about adding a dedicated security-findings section to the compaction prompt is the real fix — worth opening a follow-up issue so it isn't lost. Otherwise a malicious/oversized server can still degrade coverage, just now it's *visible*.
- The warning message is in Chinese (`扫描可能不完整`). Given the rest of the codebase mixes languages, that's acceptable, but an English log line may help non-ZH operators during triage.

Nice defensive improvement. Good to merge after the `scanNote` key-consistency tweak if you agree.

valleyxht:
Thanks for the quick and thoughtful review! Both points applied in ea627bd:

- scanNote is now always present in result_meta. Complete scans explicitly carry the value complete, with setdefault semantics so existing custom values are not overwritten. This matches the SARIF output behavior.
- The warning log is now in English so operators who do not read Chinese can triage more easily. It says the audit produced no output, and that empty results should not be interpreted as no vulnerabilities found.

I will open an issue proposing a dedicated security findings section in the compaction template shortly.

Unit tests updated accordingly. Thanks again!

boy-hack:
Thanks for applying both points, @valleyxht — `scanNote` is now always present (`setdefault` semantics, so custom values survive) and the warning log is in English, which makes triage much easier for non-ZH operators. The new tests cover the empty/whitespace/None cases plus the `setdefault` non-overwrite path. And opening a follow-up issue for the compaction-prompt security-findings section is the right call to close the root cause. LGTM — good to merge.

PR #628, opened 2026-09-01
fix(mcp-scan): preserve security findings in context compaction prompt

### LINKED ISSUE #627: mcp-scan: preserve discovered security findings across context compaction
I observed that when scanning large MCP projects containing files over 100KB, the audit agent context keeps growing until history compaction triggers. The compaction template has no dedicated section for security findings, so the LLM summarization drops the vulnerabilities already confirmed or suspected before the compaction point. The compacted summary then replaces the earlier conversation, so the agent loses track of what it already analyzed: it re-reads files it has already analyzed and may end up returning empty output. This empty result flows into calc_mcp_score([]) and gets published as a score of 100 for no vulnerabilities. A malicious project may deliberately include oversized files to evade the audit this way.

I think scans that produced no output should be detected and marked as possibly incomplete, so failed scans are no longer misreported as 100. This has already been submitted as #623.

I also suggest making the compaction process preserve findings, by extending compact.md so the summarizer keeps findings at compaction:

1. Add a top priority requirement at the prompt level: confirmed and suspected security findings (file and line numbers, risk type, evidence summary, confidence) are the final deliverable and must be preserved.
2. Add a mandatory security_findings field to the output structure, right after current_focus.
3. Add already discovered findings to the must keep list.

rwinkelman:
## Summary
- Extends `mcp-scan` `compact.md` so history compaction treats confirmed/suspected vulnerabilities as the primary deliverable.
- Adds a mandatory `<security_findings>` section (file/line, risk type, evidence, confidence) immediately after `<current_focus>`.
- Adds a regression test ensuring the compaction template exposes the security-findings section.

Fixes #627. Complements the incomplete-scan guard in #623 / PR #623.

## Context
Large MCP projects can trigger context compaction; without a dedicated findings section the summarizer drops vulnerabilities already discovered, leading to empty output and a spurious score of 100 via `calc_mcp_score([])`.

## Test plan
- [x] `pytest mcp-scan/pytests/test_compact_prompt.py`


rwinkelman:
Grooming pass: compaction prompt preserves security findings (#627). Ping if anything else is needed before merge.

boy-hack:
Thanks @rwinkelman — good catch tying #627 to the compaction pipeline, and the regression test asserting the `<security_findings>` section exists in the mcp-scan template is the right guard.

One thing worth flagging before merge: the false-negative class this fixes isn't mcp-scan-specific. There are three `compact.md` templates in the repo — `mcp-scan/mcp_scan/prompt/compact.md` (patched here), `agent-scan/agent_scan/prompt/system/compact.md`, and `skill-scan/skill_scan/prompt/compact.md`. The latter two still use the original prompt with no security-findings section, so after context compaction an agent-scan or skill-scan over a large target can still drop accumulated findings and flow into the same `calc_mcp_score([]) -> 100` false-negative. The regression test only loads the mcp-scan template, so the other two stay unguarded.

Suggestions:
- Extend the security-findings section to the agent-scan and skill-scan `compact.md` templates too (they share the identical structure), so the mitigation covers all three scanners.
- Or, if you'd rather keep this PR scoped, open follow-up issues for the other two templates so the gap isn't lost — and consider a test that asserts the section exists across *all* compact templates, not just mcp-scan.

The mcp-scan change itself is correct and I'm happy with it. The only ask is deciding whether to widen the fix or track the other two separately. Non-blocking otherwise.

PR #638, opened 2026-09-04
Add GLM-5.3-Flash fingerprint baseline

kexinoh:
## Summary
- Add a GLM-5.3-Flash random-number fingerprint baseline to `services/api_checker/baselines.json`.
- The baseline contains 500 valid samples collected with `noThink: true` using the existing API checker fingerprint workflow.

## Validation
- `iterations`: 500
- `rawData` length: 500
- `counts` sum: 500
- `python -m json.tool services/api_checker/baselines.json`
- `git diff --check`


boy-hack (approved):
Thanks @kexinoh — the GLM-5.3-Flash baseline is consistent with the existing fingerprint format: `iterations`/`rawData`/`counts` all sum to 500, `stats` look sane (mode 217, 275/500 samples), and the test counters in `test_algorithms.py` / `test_server.py` are correctly bumped from 40 → 41. Collected with `noThink: true` via the standard workflow, so this matches the other entries.

One small thing to double-check: the `model` field here is `"glm-5.3-flash"` while the existing entry is `"z-ai/glm-5.3"`. That's fine if GLM-5.3-Flash is served under a different id than GLM-5.3, but if they share a provider namespace it may be worth aligning the id convention so the detectable-models list stays predictable. Not a blocker.

LGTM — approving.

PR #648, opened 2026-09-12
fix(skill-scan): recover missing tool-call args during discovery

### LINKED ISSUE #629: 更新版本后，多次出现列目录失败导致skill扫描失败问题，上机后查看目录正常
项目信息收集报告
一、概述
本次任务旨在对位于 /app/uploads/tmp-xxx 的项目目录进行全面的信息收集与分析，以便为后续的安全评估或代码审计提供基础数据支持。

二、执行过程
在本次信息收集过程中，工具调用的执行遇到了持续性的技术阻碍。具体表现如下：

尝试使用 execute_shell 工具执行目录枚举命令（如 ls、find 等）时，系统多次反馈缺少必需的位置参数 command，尽管已尝试多种参数传递格式。
尝试使用 think 工具记录分析过程时，同样遭遇持续的参数缺失错误，提示缺少 thought 参数。
经过多轮尝试，受限于工具调用层面的异常，未能成功获取到目标目录内的文件列表及项目结构信息。
三、收集结果
由于工具调用异常未能解决，本次未能成功获取到目标项目的以下关键信息：

项目文件结构
源代码文件清单
配置文件内容
依赖关系信息
潜在的敏感数据
四、总结与建议
本次信息收集任务未能达到预期目标，主要受限于工具调用的持续性异常。建议采取以下措施以推进后续工作：

检查并修复工具调用的参数解析机制，确保参数能够被正确识别和传递。
在工具环境恢复正常后，重新执行针对该目录的信息收集流程。

qwen3-7、glm5-2均复现该问题

li872:
## Summary

Skill-scan info collection (AIG three-stage mode) often finished with an empty report after models such as qwen3-7 / glm5-2 / qwen3.7-plus emitted tool calls that the parser only partially understood. The dispatcher then called tool_func(**{}), which raised TypeError: missing 1 required positional argument: 'thought' / 'command' / 'path'. The agent treated the tools as broken, skipped the real project tree, and finished with 0 findings.

**Root cause (parser + dispatcher, not the filesystem):**

1. **Parser** (skill_scan/utils/parse.py) only extracted parameter tags inside function tags. Qwen/GLM-style JSON bodies, child tags, inline attributes, leftover free text, and tool_call wrappers produced a tool name with an empty args dict.
2. **Dispatcher** (skill_scan/tools/dispatcher.py) forwarded those kwargs unchanged. Missing required Python parameters became a raw TypeError. Legacy execute_shell (from mcp-scan prompts / model memory) is not a skill-scan tool, so listing attempts never recovered to dir_tree/ls.
3. **Stage 1** did not inject the on-disk repo tree (stage 2 already did), so a failed discovery tool sequence had no fallback file listing.

**Fix (minimal, recoverable):**

- Resilient parsing of the issue-shaped payloads plus common Qwen/GLM wrappers.
- call_compat normalizes aliases (path to file_path, cmd to command, list_dir to ls), defaults dir_tree/ls path to the uploaded project folder, remaps listing execute_shell calls onto dir_tree, and otherwise returns a retryable error that names the missing argument and shows the correct function/parameter format.
- Info Collection now injects the real directory tree as a safety net.

No shell execution was added. Arbitrary execute_shell commands are rejected with a clear “use dir_tree/ls/read_file/grep” error.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Fingerprint rule (data/fingerprints/)
- [ ] Vulnerability rule (data/vuln/)
- [ ] MCP security plugin (data/mcp/)
- [ ] Jailbreak evaluation dataset (data/eval/)
- [ ] Documentation
- [ ] Other (please describe)

## How was this tested?

Run:

    cd skill-scan
    python3 -m pytest pytests/test_tool_call_resilience.py -v
    python3 -m pytest pytests/ -v

Result: 47 passed.

Regression tests feed the #629 shapes (missing thought, missing command, dir_tree with no path / child-tag path / Qwen tool_call JSON) and assert the scanner recovers a real SKILL.md tree or returns a retryable error — never a TypeError / empty discovery.

## Tests

- skill-scan/pytests/test_tool_call_resilience.py (24 new cases)
- Full skill-scan/pytests/ suite: 47 passed

## Demo

Passing regression run:

https://raw.githubusercontent.com/li872/AI-Infra-Guard/cursor/fix-skill-scan-tool-args-a83f/docs/pr-629-demo/pytest-issue-629.log

## Related issue

Fixes #629

## Checklist

- [x] go build ./... succeeds locally (Go sources unchanged)


boy-hack:
Thanks for the thorough fix and the detailed root-cause analysis — this is a solid, well-tested contribution. The resilient parser + `call_compat` normalization layer cleanly addresses issue #629 (empty discovery reports when Qwen/GLM-class models emit malformed or legacy-named tool calls), and the regression suite (24 new cases, full suite green) gives good confidence. I left a few suggestions below, mostly around edge cases and a pre-existing limitation — none are blockers. **Strengths** - Clear separation of concerns: parser (`parse.py`) extracts args from many shapes (inline attrs, child tags, JSON bodies, key=value lines, leftover text); `call_compat.py` normalizes aliases and returns retryable errors instead of raw `TypeError`. - No shell execution added; `execute_shell`/legacy shell tools are explicitly rejected with a clear pointer to `dir_tree`/`ls`/`read_file`/`grep`. Good security posture. - The stage-1 `inject_repo_tree=True` safety net is a nice fallback so a failed discovery sequence still has a real file listing. **Suggestions / things to consider** 1. **Raw free-text → first required arg.** In `apply_arg_aliases`, leftover body text (`RAW_BODY_ARG`) is mapped onto the *first* required parameter. For `think` that is correct, but for a tool whose first required arg is `command`/`query`/`pattern`, arbitrary model text could be coerced into that argument. Consider limiting the raw-body fallback to text-like params (e.g. `thought`, `content`, `query`) rather than positional-first, so we never silently turn a stray sentence into a `command`. 2. **`_looks_like_path` heuristic.** It accepts tokens matching `^[\w.:/\\-]+$` containing `/` or `\`. A long instruction string containing a slash (e.g. "read the file at /tmp/x") could be misclassified as a path for `dir_tree`. The risk is bounded (it only fills `path`/`file_path` when missing), but tightening the heuristic (reject whitespace/newlines already done — good) or requiring an absolute/relative prefix could reduce false positives. 3. **Multi-tool-call messages.** `parse_tool_invocations` still returns only `invocations[0]`, so a single model message emitting several `<function>` blocks will only execute the first. This is pre-existing behavior, not introduced here, but worth tracking as a follow-up since resilient parsing now makes multi-call recovery more realistic. 4. **`canonicalize_tool_name` quote stripping** is reasonable, but note a tool literally named `bash`/`shell`/`run` will be unconditionally routed to `remap_legacy_shell` (good — they are rejected). Just confirming that is intended and not a real skill-scan tool name collision. 5. **Minor:** the `<recovered>` wrapper condition `args.get("_recovered_from")` is effectively redundant with `tool_name != canonical` (the `_recovered_from` key is popped from the prepared copy before execution), but it is harmless. Overall this looks good to me. Once CI is green and the raw-body fallback in (1) is tightened to avoid coercing free text into non-text params, this is ready to merge. Nice work on the regression tests — they make the recovery behavior easy to trust.

li872:
@boy-hack Thanks for the review.

Addressed (1): leftover `_raw_body` text is now only mapped onto text-like params (`thought` / `content` / `text`, plus `query`). It is never coerced into `command`, `pattern`, or other non-text args — those still return the retryable missing-argument error.

Also tightened `_looks_like_path` so instruction strings that merely contain a slash (e.g. `read the file at /tmp/x`) are not treated as paths, and dropped the redundant `_recovered_from` wrapper check.

Left (3) as a follow-up (pre-existing first-invocation-only behavior) and confirmed (4) that bash/shell/run remapping is intended.

Regression tests updated; full `skill-scan/pytests` suite is green (`5487462`).

boy-hack:
@li872 Thanks for the quick turnaround — I pulled the updated diff and verified each change against the code:

- **(1) Raw-body → text-only params:** confirmed. `_RAW_BODY_TEXT_PARAMS = frozenset({"thought", "content", "text", "query"})` and `apply_arg_aliases` only maps leftover `_raw_body` text onto those (or onto `path`/`file_path` when the token genuinely `_looks_like_path`). `command`/`pattern` never get coerced — they still return the retryable missing-argument error. Exactly what I asked for.
- **(2) Tightened `_looks_like_path`:** confirmed. The new `_EXPLICIT_PATH_RE` (`^(?:~/|/|\./|\.\./|[A-Za-z]:[\\/])…`) plus `_COMPACT_REL_PATH_RE` (short, slash-separated, no spaces) means "read the file at /tmp/x" is no longer treated as a path. Good hardening.
- **(4) bash/shell/run remapping:** confirmed intended — `_LEGACY_SHELL_TOOLS` routes them to `remap_legacy_shell` and they're rejected with a clear pointer to `dir_tree`/`ls`. No real skill-scan tool collision.
- **(5) Redundant `<recovered>` wrapper check:** confirmed removed — `prepare_tool_call` now pops `_recovered_from` before execution and the dispatcher only adds the wrapper when `tool_name != canonical`.

The `parse.py` resilience (inline attrs, child tags, `<arg_key>/<arg_value>`, JSON bodies, key=value lines, `<tool_call>` wrappers) and the stage-1 `inject_repo_tree=True` safety net look correct, and the security posture is unchanged (no shell execution, legacy shell tools rejected).

**One thing to keep on the radar (follow-up, not a blocker):** `parse_tool_invocations` still returns only `invocations[0]`, so a single model message emitting multiple `<function>` blocks still executes just the first. As you noted this is pre-existing, and the resilient parser now makes multi-call recovery realistic — worth a dedicated follow-up PR so a multi-tool message doesn't silently drop the 2nd+ call.

**Test coverage note:** the new `test_tool_call_resilience.py` (24 cases) covers the issue-#629 shapes and the full suite is green at `5487462` — that gives good confidence the recovery path is deterministic. If CI is green on this head, this is ready to merge from my side. Nice work.

li872:
@boy-hack Thanks for re-checking — glad it's ready once CI is green. I'll leave the multi-tool `invocations[0]` follow-up for a dedicated PR after this merges.

boy-hack:
@li872 Confirmed on the head at `5487462` — I pulled the actual `call_compat.py` and verified each point:

- **(1)** `_RAW_BODY_TEXT_PARAMS = frozenset({"thought", "content", "text", "query"})`, and `apply_arg_aliases` only maps leftover `RAW_BODY_ARG` onto those (or onto `path`/`file_path` when the token genuinely `_looks_like_path`). `command`/`cmd`/shell never get coerced — they still hit the retryable missing-arg error. ✅
- **(2)** `_EXPLICIT_PATH_RE` (`^(?:~/|/|\./|\.\./|[A-Za-z]:[\\/])…`) + `_COMPACT_REL_PATH_RE` (short, slash-separated, no spaces, `<=128` chars) means "read the file at /tmp/x" is no longer treated as a path. ✅
- **(5)** the `<recovered>` wrapper no longer relies on the redundant `_recovered_from` check — `prepare_tool_call` pops it before execution. ✅

Good call leaving **(3)** (multi-tool `invocations[0]` only) as a dedicated follow-up PR — the resilient parser now makes that recovery realistic, and a single message emitting multiple `<function>` blocks shouldn't silently drop the 2nd+. 

From my side this is ready to merge once CI is green on `5487462`. Thanks for the quick, precise turnaround and for not over-engineering the recoverable layer. 👍

PR #649, opened 2026-09-14
fix(scanners): preserve findings during context compaction

Beverly621:
## Summary

- Preserve confirmed and suspected security findings when `agent-scan` or `skill-scan` compacts its conversation history.
- Align both compaction prompts with the behavior already merged for `mcp-scan` in #628, following the maintainer's [review note](https://github.com/Tencent/AI-Infra-Guard/pull/628#issuecomment-5503309753).
- Add focused regression tests that require the `<security_findings>` section and its preservation priority in both scanners.

This is a narrow follow-up: it changes prompt templates and regression coverage only, with no scanner control-flow or API changes.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Fingerprint rule (`data/fingerprints/`)
- [ ] Vulnerability rule (`data/vuln/`)
- [ ] MCP security plugin (`data/mcp/`)
- [ ] Jailbreak evaluation dataset (`data/eval/`)
- [ ] Documentation
- [ ] Other (please describe)

## How was this tested?

- [x] `cd agent-scan && python3 -m pytest pytests/test_compact_prompt.py -q` — 1 passed.
- [x] `cd skill-scan && python3 -m pytest pytests/test_compact_prompt.py -q` — 1 passed.
- [x] `cd mcp-scan && python3 -m pytest pytests/test_compact_prompt.py -q` — 1 passed, confirming consistency with the existing implementation.
- [x] `go build ./...` — passed.
- [x] `git diff --check origin/main...HEAD` — passed.

## Related issue

Follow-up to #628 and the maintainer review linked above. No new issue was opened.

## Checklist

- [x] `go build ./...` succeeds locally

<details>
<summary>中文说明</summary>

## 概要

- 修复 `agent-scan` 与 `skill-scan` 在对话历史压缩时可能丢失已确认或疑似安全发现的问题。
- 将两个扫描器的压缩提示与 #628 已合并的 `mcp-scan` 行为保持一致；该补充范围来自维护者在 #628 中的 Review 建议。
- 分别增加回归测试，要求压缩结果包含 `<security_findings>`，并将安全发现设为不可丢弃的最高优先级内容。

本 PR 仅修改两个提示模板及其回归测试，不修改扫描器控制流程、API 或其他业务逻辑。

## 验证

- `agent-scan` 目标回归测试：1 passed。
- `skill-scan` 目标回归测试：1 passed。
- `mcp-scan` 现有一致性测试：1 passed。
- `go build ./...`：通过。
- `git diff --check origin/main...HEAD`：通过。

</details>


boy-hack:
Thanks for this — it's a clean, low-risk follow-up that brings `agent-scan` and `skill-scan` in line with the `mcp-scan` compaction behavior already merged in #628, and the regression tests assert the `<security_findings>` section plus the "最高优先级 / 不可丢弃" priority text, so a future regression will be caught. A few small notes:

**1. Fix the path comparison in the test**
`assert "不可丢弃" in template or "最高优先级" in template` — the prompt uses **`最高优先级`** literally (the "最高优先级（不可丢弃）" heading), not the two characters `不可丢弃` as a standalone substring. The `or` makes the assertion pass (because `最高优先级` is present), so it works today, but the intent reads as if both were checked. Consider asserting `"最高优先级" in template and "不可丢弃" in template` (both strings do appear) so the test actually pins the priority + the "do-not-drop" requirement rather than passing on the weaker condition.

**2. `security_findings` priority is good, but consider also preserving *confirmed findings* even when outside the compact window**
The prompt now instructs the model to never drop findings, which is the right fix for the loss-during-compaction bug. One residual gap (pre-existing, not introduced here): if a finding was surfaced in an *earlier* compacted turn and the model paraphrases it, the structured `file/line/risk_type/evidence/confidence` fields can still drift. Not a blocker for this prompt-only PR — just noting it as a follow-up, since the safety benefit depends entirely on the model faithfully re-emitting the section.

**3. Minor: the `mcp-scan` consistency test passes unchanged**
You mention `mcp-scan`'s existing `test_compact_prompt.py` still passes, confirming parity — good. No action needed, just acknowledging the consistency check.

This is a docs/prompt-only change with no control-flow or API impact, tests are green, and `go build ./...` passes. Once the assertion in (1) is tightened (so it actually guards both invariants), this is ready to merge.

Beverly621:
Thanks for the review. I tightened both regression tests in `22bf80b` so they now independently require the `最高优先级` and `不可丢弃` invariants.

Validation:
- `agent-scan`: 1 passed
- `skill-scan`: 1 passed
- `git diff --check`: passed

I’ve kept the structured-finding drift concern from point 2 out of this prompt-only PR, as suggested, and will treat it as a separate follow-up.

boy-hack:
@Beverly621 Confirmed at `22bf80b` — both regression tests now independently assert `"最高优先级" in template` **and** `"不可丢弃" in template` (plus `<security_findings>` and `安全发现`), so the test actually pins the priority + "do-not-drop" invariants instead of passing on the weaker `or` condition. ✅

Agreed on keeping the structured-finding drift concern (point 2) out of this prompt-only PR and tracking it separately — the safety benefit here is the model faithfully re-emitting the `<security_findings>` section, and that's now enforced by the tests.

This is a clean, low-risk follow-up that brings `agent-scan`/`skill-scan` in line with the `mcp-scan` compaction behavior from #628, with `go build` and `git diff --check` passing. From my side this is ready to merge. Nice work. ✅

PR #651, opened 2026-09-14
fix(mcp-scan): rate-limit aware retry for LLM calls

toBeYoungD:
Internal/shared LLM gateways commonly enforce QPM limits. mcp-scan's LLM calls had no retry or backoff: a single 429 or transient gateway error propagated up and killed the whole scan task (main agent loop via utils/llm.py chat_stream, and redteam attacker/evaluator/target calls via AsyncOpenAI).

Add mcp_scan/utils/llm_retry.py with shared sync/async retry wrappers (same semantics as the prompt-eval fix in PR #637):
- detect rate-limit errors via status code 429 and message markers (rate limit / too many requests / quota exceeded / resource exhausted)
- honor Retry-After when present (capped at 60s)
- otherwise back off 8s/16s/32s (capped at 60s); non-rate errors keep a short exponential backoff (1s base, capped at 8s)
- max 5 trials, re-raise the last error when exhausted

Apply the wrappers to all four LLM call sites. LLM.chat() keeps its existing empty-response retry loop unchanged; when retries are exhausted it now returns an error placeholder instead of crashing the whole scan task.

Tests: pytests/test_llm_retry.py (15 cases) cover rate-limit detection, Retry-After extraction, the backoff schedule, and sync/async recovery/exhaustion paths. All pass with Python 3.12 in the agent image; existing test_mcp_sdk_v2 passes unchanged.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Fingerprint rule (`data/fingerprints/`)
- [ ] Vulnerability rule (`data/vuln/`)
- [ ] MCP security plugin (`data/mcp/`)
- [ ] Jailbreak evaluation dataset (`data/eval/`)
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [ ] `go build ./...` succeeds locally
- [ ] I followed the existing file/rule format conventions for this contribution type
- [ ] I have not introduced hardcoded secrets or credentials


boy-hack:
Thanks for this — it's a well-scoped fix for a real problem (a single 429 or transient gateway error used to abort the whole scan), and the shared `llm_retry.py` wrapper matches the semantics already used in #637, which keeps behavior consistent across scanners. The test coverage (detection, `Retry-After` extraction, backoff schedule, sync/async recovery + exhaustion) is solid. A few things worth tightening before merge:

**1. Potential `NameError` on the retry-exhausted path in `llm.py`**
In `LLM.chat`, when `call_with_retry` re-raises after the last trial you do:
```python
except Exception as e:
    logger.error(f"LLM chat failed after retries: {type(e).__name__}: {e}")
    ret = "连接LLM失败，已重试多次…"
    break
if ret != "":
    break
else:
    ...
    if ret_usage:
        return ret, usage   # <-- usage is unbound here
```
`usage` is only ever assigned inside the `while` loop's `ret, usage = self.chat_stream(message)`. On the exception path `usage` is never bound, so any caller passing `ret_usage=True` will raise `NameError` instead of returning `(ERROR_PLACEHOLDER, None)`. The error placeholder return at the top of the loop (`retry > 5`, empty output) has the same issue. Recommend initializing `usage = None` before the loop (or returning `ret, None` in the except branch) — and adding a quick `ret_usage` regression test so this doesn't regress.

**2. `chat_stream` (empty-response) retry loop is now nested under rate-limit retry**
Previously `chat_stream` itself ran the 1.3s/×5 empty-output retry. Now `call_with_retry` wraps `chat_stream`, so a fully-successful-but-empty response still loops through the 1.3s empty-retry *inside* `chat_stream`, then `call_with_retry` sees a normal (non-rate-limit) return and stops. That's fine functionally, but note the empty-output path no longer gets rate-limit retries of its own — acceptable, just flagging the interaction so it's intentional.

**3. Minor: `_RATE_LIMIT_BACKOFF` table vs the docstring**
Docstring says "8s/16s/32s (capped at 60s)" and the tuple is `(8.0, 16.0, 32.0, 60.0, 60.0)` — 5 entries for 5 trials, good. But `backoff_delay` indexes by `attempt` (0-based) and clamps with `min(attempt, len-1)`, so trial 4 and 5 both wait 60s. That matches the comment, just confirming the 60s plateau is intended rather than, say, a longer final wait.

**4. Small robustness nit**
`is_rate_limit_error` checks `str(exc).lower()` plus a `"429"` substring marker. Some SDKs raise a generic `APIConnectionError` whose message doesn't contain those tokens but *does* carry `status_code`. Since `status_code` is only checked for `== 429`, a 429 wrapped in a non-API-status exception class would be missed and fall into the 1s→8s non-rate backoff. Low impact (it still retries), but you might widen the status check to `getattr(exc, "status_code", None) in (429, 42900)` or treat any 4xx/5xx with a `Retry-After` as rate-limited.

None of these are blockers and the security posture is unchanged. If you address (1) — the unbound `usage` is a real correctness bug for `ret_usage` callers (and the existing `test_mcp_sdk_v2` suite may not exercise `ret_usage=True`, which is exactly why it passed) — this is ready to merge. Nice work on the symmetry with #637.

boy-hack:
Correction on point (1) above — I misread the diff hunk. The `usage = None` initialization sits *above* the changed block (it's preserved, not removed), so `usage` stays bound on the retry-exhausted path and `return ret, usage` for `ret_usage=True` callers safely returns `(placeholder, None)` rather than raising `NameError`. Apologies for the false alarm — that's not a bug. The remaining notes (points 2–4) stand, and none are blockers.

toBeYoungD:
Thanks for the thorough review — and to @boy-hack for the correction on (1); happy to hear the init was preserved.

On (2) and (3): both intentional — empty responses keep the original 1.3s × 5 loop (an empty completion isn't rate limiting), and the 60s plateau after trial 4 is by design to keep worst-case waits bounded.

On (4): adopted in 03e29729 — is_rate_limit_error now also accepts status_code in (429, 42900) and treats any error carrying a Retry-After header as rate-limited (covering wrapped 429s and gateway 5xx backoff signals). Added three detection tests for those cases; all 18 pass.

boy-hack:
@toBeYoungD Thanks for closing the loop — verified the `is_rate_limit_error` change against `llm_retry.py` at head:

**4. Detection widening — fixed and verified.** `is_rate_limit_error` now returns `True` when `status_code in (429, 42900)` **or** a `Retry-After` header is present (via `retry_after_seconds`), before falling back to the message-marker match. This correctly catches 429s wrapped in a non-API-status exception class and treats gateway 5xx-with-`Retry-After` as rate-limited. The three new detection tests (wrapped `429`, `42900`, and header-only) are in `test_llm_retry.py`, bringing it to 18 passing. ✅

**2 & 3 — confirmed intentional.** Empty completions keep the original 1.3s×5 loop (empty output ≠ rate limiting), and the 60s plateau after trial 4 is a deliberate worst-case bound. Both make sense. ✅

From a review standpoint this is ready to merge — the shared `llm_retry.py` keeps mcp-scan consistent with the prompt-eval fix in #637, and the four call sites now degrade gracefully instead of aborting the whole scan.

One small forward-looking nit (non-blocking, for whenever you touch this file again): `call_with_retry`/`acall_with_retry` log `logger.error` on *every* retry attempt, so under a sustained QPM-limit window a single scan will emit several `error`-level lines before each backoff. Since a retry is expected behavior on a 429, `logger.warning` (or collapsing to a single throttled log) would keep the error stream cleaner without losing signal. Totally optional — leaving it as-is is fine too.

PR #653, opened 2026-09-15
feat(deepteam): write finished test cases to disk during a run

helo060228:
## Summary

A red team run with many techniques and datasets can take hours, and every finished test case used to live in memory until the very end, so an interrupted run lost all of the work done so far. This writes each finished `RedTeamingTestCase` to disk while the run is still going.

- New `AIG-PromptSecurity/deepteam/red_teamer/progress_log.py` with `ProgressLog`: appends JSONL records and flushes every `flush_every` cases with `flush()` + `os.fsync()`. Any abrupt end of the process costs at most `flush_every - 1` cases instead of everything.
- `path=None` turns the log into a no-op object, so that is how the feature is disabled and callers never check for `None` themselves.
- Wired into both flows: the sync `red_team()` (all five append paths, including the `ignore_errors` ones) and the async path (a case is written as soon as its task completes, not when the whole vulnerability type finishes).
- Enabled by default at `logs/red_team_progress.jsonl`, following the existing `logs/` convention used by `save_risk_assessment_report()`. `--progress-log <path>` in `cli_run.py` changes the file; passing an empty value disables it.
- `ProgressLog` and its defaults are exported from `deepteam.red_teamer`.

The "断点续跑" (resume) half of the issue is intentionally **not** in this PR: `CONTRIBUTING.md` asks for one logical change per PR, and the JSONL file is already shaped to be consumed as a resume checkpoint in a follow-up.

## Type of change

- [x] New feature

## How was this tested?

- `python -m pytest tests/test_progress_log.py` — 8 new tests: the flush threshold (nothing is written before it is reached), the remainder written by `close()`, an abrupt end that keeps what was already flushed, appending across runs, the JSON aliases of the written records, the `path=None` no-op, directory creation, and `flush_every < 1`.
- `python -m pytest tests/test_deepteam.py` — the existing smoke test still passes.
- Manual end-to-end check: with `flush_every=2`, recording 5 cases leaves 4 lines on disk, i.e. exactly the documented bound for a process that dies before `close()` runs.

## Notes for reviewers

- The sync flow has five places that do `red_teaming_test_cases.append(...)` followed by `progress_log.record(...)`. I kept that repeated on purpose instead of hiding it in a list subclass or a closure: an explicit second line is easier to follow than a container with side effects, and it keeps every write point visible in the diff.
- The same five paths are the `ignore_errors` branches. They also record, so a run that is degraded by rate limits still leaves a partial result behind.

Two pre-existing things I ran into while testing, both untouched here:

- `deepteam/telemetry.py` imports `sentry_sdk`, but it is missing from `AIG-PromptSecurity/requirements.txt`, so `python -m pytest tests/` cannot collect anything on a clean checkout until it is installed (the existing `tests/test_deepteam.py` fails the same way). Happy to fix it in this PR or in a separate one — your call.
- The root `.gitignore` has a bare `tests` entry, so new files under `tests/` need `git add -f` to be picked up.

## Related issue

Refs #537 — this PR implements the "日志即时写入" half, "断点续跑" will follow up.

## Checklist

- [ ] `go build ./...` succeeds locally — not applicable to this Python-only change; `AIG-PromptSecurity` was smoke-run instead (`from deepteam import red_team`, plus importing the new module)


boy-hack:
Thanks for this — it's a clean, well-scoped contribution and the design is sound. Writing each finished `RedTeamingTestCase` to JSONL as the run progresses (with `flush()` + `os.fsync()` every `flush_every`, and `close()` catching the remainder) directly solves the "hours of work lost on interrupt" problem, and the append-mode file is already shaped as a resume checkpoint for the follow-up. Good separation of the "log" half from "resume" half per `CONTRIBUTING.md`. The 8 new tests (flush threshold, abrupt-end bound, cross-run append, JSON aliases, `None` no-op, dir creation, `flush_every < 1`) cover the contract well, and keeping the explicit `progress_log.record(...)` at each of the five append points is the right call for readability.

A few points, two of which are real (non-blocking) bugs:

**1. "Pass empty value to disable" is broken (correctness bug).**
The PR description and the `--progress-log` help text say *"pass an empty value to disable it"*, but `ProgressLog` only guards on `is None`:
```python
def record(self, test_case):
    if self.path is None:   # "" (empty string) is NOT None
        return
    ...
    with open(self.path, "a", ...) as f:   # open("") -> FileNotFoundError
```
If a user runs `--progress-log ""`, `args.progress_log` is `""`, which flows into `RedTeamer(progress_log_path="")` → `ProgressLog(path="")`, and the first `record()` call does `open("", "a")` and crashes. The library-level `None` path works, but the CLI's documented "empty disables" contract doesn't. Minimal fix in `cli_run.py`:
```python
run_red_team(..., progress_log_path=args.progress_log or None)
```
(or guard with `if not self.path:` in `ProgressLog` — but the explicit `None`-means-disabled contract is clearer). Please keep the two consistent.

**2. `close()` isn't called on exception (minor data-loss edge).**
`progress_log.close()` runs only after the `pbar.close()` at the end of `red_team`/`a_red_team`. If the run raises (rate-limit exhaustion, target error with `ignore_errors=False`, etc.) before that point, the last `< flush_every` cases still in the buffer are lost — beyond the documented `flush_every - 1` bound. Wrapping the body in `try/finally: progress_log.close()` would tighten that to the stated guarantee. Not a blocker, just more robust.

**3. `os.makedirs` ignores the `exist_ok`/`mode` and runs even when `path` looks like a bare filename — fine, but note** `directory = os.path.dirname(path or '')` is skipped for a bare filename (e.g. `progress.jsonl`), writing into cwd. Expected, just flagging so the default `logs/red_team_progress.jsonl` (which creates `logs/`) is the intended shape.

**4. Pre-existing `sentry_sdk` gap (out of scope, your call).**
You noted `deepteam/telemetry.py` imports `sentry_sdk` but it's missing from `AIG-PromptSecurity/requirements.txt`, so `pytest` can't collect on a clean checkout. Since this PR adds a test file, it'd be a small courtesy to add `sentry_sdk` to `requirements.txt` here (or open a separate issue). Left to you.

**5. `.gitignore` `tests` entry.** The diff does include `AIG-PromptSecurity/tests/test_progress_log.py` as added, so it's tracked — just confirming it was force-added and won't be silently dropped by the bare `tests` ignore rule on push.

Net: the design and tests are good. Once **(1)** is fixed (empty-string → `None`) and **(2)** is considered, this is ready to merge. Nice work keeping the write points explicit and the resume format forward-compatible. 🚀

helo060228:
Thanks for the careful read.

(1) is a regression I introduced myself. The first version guarded `if not self.progress_log_path`, and when I refactored to the no-op object I narrowed that to `is None`, so `""` reached `open("")`. Fixed at both ends now — `args.progress_log or None` in `cli_run.py` and `if not self.path:` in `ProgressLog` — plus a test for the empty string, which is exactly the case my `None`-only test missed.

(2) Done: both flows wrap the run in `try/finally` and close there, so the buffered cases no longer wait for the end of the run. That change is a re-indent of the two loop bodies, nothing else.

(3) Right, a bare filename writes into the working directory by design, and the default stays `logs/red_team_progress.jsonl`.

(4) Added `sentry-sdk>=2.0.0` to `AIG-PromptSecurity/requirements.txt`. Without it the test file this PR adds cannot be collected on a clean checkout at all.

(5) Confirmed: it was force-added and is tracked.

Locally after the changes: 10 passed.

boy-hack:
@helo060228 Thanks for the quick turnaround — I pulled the updated head and verified every point against the code:

**1. Empty-value disable — fixed and verified.** `cli_run.py:151` now does `progress_log_path=args.progress_log or None`, and `ProgressLog.record`/`flush` guard on `if not self.path:` (so both `None` and `""` are no-ops). The new `test_empty_path_is_a_no_op_too` covers exactly the case my `None`-only review missed. ✅

**2. `close()` on exception — fixed and verified.** Both `red_team` and `a_red_team` now wrap the loop body in `try/finally: progress_log.close()`, so a run that raises (rate-limit exhaustion, `ignore_errors=False` target error) still flushes the buffered remainder. The data-loss bound now matches the documented `flush_every - 1`. ✅

**4. `sentry-sdk>=2.0.0` added to `AIG-PromptSecurity/requirements.txt`** — so `tests/test_progress_log.py` now collects on a clean checkout. ✅ (and good catch that `test_deepteam.py` had the same latent breakage.)

**5.** Confirmed `tests/test_progress_log.py` is force-added and tracked despite the bare `tests` `.gitignore` rule.

All 10 tests passing locally, and the design is unchanged: explicit `record()` at each of the five write points, append-mode JSONL shaped as a resume checkpoint. This is ready to merge from a review standpoint. 🚀

One forward-looking note (not for this PR, just to keep #537 in mind): the JSONL records use `model_dump(mode="json", by_alias=True)`, which is exactly what a follow-up resume consumer will want — when you build the "断点续跑" half, you'll be able to `RedTeamingTestCase.model_validate` them back directly. Nice forward-compatibility.

helo060228:
Thanks for the thorough verification.

That is exactly the plan for the follow-up: the resume consumer will `RedTeamingTestCase.model_validate` the JSONL lines back, so the checkpoint needs no separate schema. I will open it as its own PR once this one is in, so it builds on the merged base rather than on an unmerged branch.

I do have a short design sketch ready, and it touches one thing worth your input before I write code: the simulated attacks live only in memory (`save_test_cases_as_simulated_attacks` sets `self.simulated_attacks` and does not persist it), so a restarted run regenerates different attacks and "skip what is already done" has nothing to match against. Resume therefore has to persist the attacks as well as the results, and merge the old results into the final report. Happy to post that sketch in #537 for your input if you want to see it before the PR.

boy-hack:
@helo060228 Sounds good — please do post the sketch in #537, I'd like to see it before you write the resume half. Quick reaction on the one open question so you can start from the right shape:

**Persist the generated attacks, not the generator.** Since `save_test_cases_as_simulated_attacks` only holds `self.simulated_attacks` in memory and the attacks are randomized per run, "skip what's done" has no stable key to match against unless you serialize the actual attack instances. So the resume checkpoint should persist the *concrete* attack objects you generated, not re-derive them. On resume, load those back and reuse them as the work list — that's what makes "already-done" a real identity instead of a coin flip.

**Stable identity for dedup.** Give each persisted attack a deterministic key (attack type + target id + a seed/hash of the realized content). Then "skip completed" = `key in {keys from JSONL results}`. A bare random seed won't survive a crash unless the seed itself is persisted, so prefer hashing the realized attack over relying on the seed.

**Merge into the final report.** Because results are already append-mode JSONL by design, the report merge is just `RedTeamingTestCase.model_validate(line)` over every line (old + new) and re-aggregate — no separate schema, consistent with the checkpoint format this PR ships.

Net: the direction is right and it composes cleanly with this PR's forward-compatible JSONL. Go ahead and open the follow-up PR off the merged base. 🚀

PR #668, opened 2026-09-20
feat(deepteam): resume an interrupted red team run

helo060228:
## Summary

The "断点续跑" half of #537, on top of the merged #653.

A run can take hours, and the attacks it evaluates are generated randomly, so a restarted process cannot tell which cases are already done. This persists the generated attacks next to the results and lets a new run continue where the previous one stopped.

- `resume.py`: `AttackCheckpoint` (attacks JSONL with a metadata header), a content-based `case_key`, and a reader for the finished cases that `progress_log` writes.
- `--resume` (off by default) reuses the checkpoint, skips the cases a previous run already finished, and merges the old results into the final report instead of dropping them.
- A checkpoint written with a different configuration is rejected instead of silently mixing two runs.
- Attack identity is a hash of the realized content rather than the random seed, as you suggested in #653: a seed cannot survive a crash on its own.

## Type of change

- [x] New feature

## How was this tested?

- `python -m pytest tests/test_resume.py tests/test_progress_log.py tests/test_deepteam.py` — 20 passed.
- The new tests cover the key stability (an attack and the test case built from it hash the same), the checkpoint round trip and half-written lines, the skip-completed filtering, the mismatch rejection, and the resume wiring on a real `RedTeamer` (with a dummy key, so it runs in CI).
- Manual end-to-end check without any model call: a checkpoint with two attacks plus one finished result → resume reads both attacks, recognises the finished one, leaves only the other pending, and keeps its `actual_output`.

Two things worth knowing:

- `actual_output` only has a `serialization_alias`, so validating a written record as-is silently drops the target model output. `load_completed_cases` maps `actualOutput` back; without that, a resumed report would show finished cases as if the model had produced nothing. Relevant to anything else that reads these files.
- `tests/test_custom_vulnerability.py` fails to collect on `main` today (`TypeError: 'module' object is not callable`). Unrelated to this PR — I reproduced it on a clean `f25e325`.

## Related issue

Refs #537 — the reporter closed it after the first half landed, this is the second half.

## Checklist

- [ ] `go build ./...` succeeds locally — not applicable, Python-only change; `AIG-PromptSecurity` was smoke-run instead


boy-hack:
Impressive — the second half of #537 is well-structured and the tests cover the tricky bits (half-written checkpoint lines, mismatch rejection, completed-case `actual_output` preservation, real `RedTeamer` wiring). The decision to persist the *realized attacks* (content-hash `case_key`) rather than the random seed, as discussed in #653, is the correct one — it's what makes a crash-recoverable resume actually stable. A few substantive points:

- **`build_metadata` omits `target_purpose`.** Today the mismatch guard rejects a resumed run only when `vulnerabilities` / `attacks` / `attacks_per_vulnerability_type` differ. But if a user interrupts scan A (target X), then kicks off scan B against a *different* target with the same vuln/attack set, `--resume` will load scan A's attacks and merge scan A's completed results into scan B's report — silent cross-target contamination. I'd strongly suggest adding `target_purpose` (or a stable hash of it) to the metadata and checking it in `ensure_checkpoint_matches`. This is the one gap that could produce a *wrong report* rather than just an error.
- **`case_key` truncation to 16 hex chars (64 bits).** Collision probability is negligible at realistic attack counts (hundreds–low thousands), so this is fine in practice, but it's not a cryptographic guarantee. If `attacks_per_vulnerability_type` is ever raised to very large values, consider a longer prefix. Not blocking.
- **Progress bar semantics on resume.** In `a_red_team`, `pbar.update(len(attacks))` counts *all* attacks including the resumed/completed ones, so the bar jumps straight to full for resumed cases while the async loop only re-evaluates the pending subset. Cosmetic only, but worth a comment or counting `pending_attacks` for the bar so the UX matches reality.
- **`_save_attacks` only runs in the fresh-generate branch.** Correct: on resume the checkpoint is loaded, not regenerated, so no clobber. Just confirm a `--resume` run that finds *no* usable checkpoint cleanly falls through to a fresh generate+save — your `None` return path handles that (verified in `test_red_teamer_ignores_checkpoint_without_resume`).

The `actual_output` alias mapping (`_RESULT_ALIASES`) is a nice catch — anything else reading these JSONL files should be aware of it.

Overall this is a strong, well-tested feature. The `target_purpose` gap is the one thing I'd want closed before merge; the rest are nits.

helo060228:
Good catch on `target_purpose` — that one produces a wrong report rather than an error, so it is fixed here rather than deferred:

- `build_metadata` now records `target_purpose` (empty string when unset) and `ensure_checkpoint_matches` checks it.
- `test_checkpoint_mismatch_is_rejected_on_target_purpose` covers the cross-target case, and I checked the wiring directly: the same target resumes, a different one is rejected with `Cannot resume: the attacks checkpoint was written with a different target_purpose`.

On the rest:

- The 16-hex key: agreed that it is not a cryptographic guarantee. I kept 64 bits because it only has to separate cases inside a single run; easy to widen if `attacks_per_vulnerability_type` is ever meant to go very large.
- Progress bar: added a comment that resumed cases count as done on purpose, since the bar is meant to show how much of the run is finished, not how much was re-evaluated.
- `_save_attacks` only in the fresh branch: confirmed intentional, and the `None` return from `_resumed_attacks` is what makes a `--resume` run without a usable checkpoint fall through to a fresh generate and save.

21 tests pass locally after the change.

boy-hack:
@helo060228 Thanks for closing the `target_purpose` gap — that was the only thing standing between this PR and a clean merge, and the cross-target test (`test_checkpoint_mismatch_is_rejected_on_target_purpose`) plus your direct wiring check cover it well. The other nits are acknowledged too: the 16-hex `case_key` collision note is non-blocking, and the progress-bar comment correctly documents that resumed cases count as done on purpose (the bar tracks overall run completion, not re-evaluation). The `actual_output` alias handling (`_RESULT_ALIASES`) remains a genuinely useful catch for anyone else reading those JSONL files.

I'm satisfied with this one — the second half of #537 is well-structured, well-tested, and the last correctness blocker is resolved. Ready to merge from my side. Nice work.

Issue #85, opened 2025-09-18
add extra_body and extra_headers support?


limike627:
The model API configuration page does not support setting additional parameters, such as extra_headers and extra_body.

In the OpenAI SDK, there is an **extra_body** parameter available in the Completions.create method, which allows users to add additional JSON properties to the request body. This feature can be very useful in various scenarios, especially when needing to send extra data that is not covered by the predefined parameters.

Besides, I often use OpenRouter api and need to configure **the provider router** along with specific **quantization** settings to reduce model size and achieve the best model performance.

Please consider adding this feature.

zonashi:
Thank you for your suggestion. We will add this feature in the next version.

Beverly621:
> Thank you for the suggestion! Here's an update on this request.
> 
> ## Current Status
> The current `OpenAI` client in AI-Infra-Guard uses the official `openai-go` SDK and supports `BaseUrl`, `ApiKey`, and `Model` configuration. **`extra_body` and `extra_headers` are not yet exposed as configurable options in the UI.**
> 
> ## Workaround for Now
> If you need to pass custom headers (e.g., for OpenRouter provider routing), you can use the **custom HTTP headers** feature already available in the scan configuration:
> 
> {
>   "token": "your-api-key",
>   "base_url": "https://openrouter.ai/api/v1/",
>   "model": "openai/gpt-4o",
>   "headers": {
>     "HTTP-Referer": "https://your-app.com",
>     "X-Title": "Your App",
>     "X-Quantization": "int4"
>   }
> }
> The `headers` field in the model configuration already allows passing arbitrary HTTP headers, which covers most OpenRouter-specific routing use cases.
> 
> ## Feature Roadmap
> For `extra_body` support (e.g., adding custom JSON properties to the request body like `provider.order` or `provider.quantizations`), we have this on our roadmap. It requires extending the model config schema and the underlying SDK call. We will update this issue when it's implemented.
> 
> If you have specific use cases beyond header-based routing, please share them here — it will help us prioritize!

😃😃😃 Hi maintainers, I’d like to help with this issue.

Since custom headers already have an existing configuration path, I propose limiting the initial PR to `extra_body` support:

- Add an optional JSON-object field to the model configuration.
- Preserve it through model create, update, and read operations.
- Forward it to model validation and scanner LLM requests.
- Keep existing model configurations fully backward-compatible.
- Add request-capture regression tests to verify that provider-specific fields reach the outgoing request body.
[...]

PR #669, opened 2026-09-20
feat(models): send extra headers and body fields to the model API

helo060228:
## Summary

Implements the backend half of #85: the model configuration could set a base URL, a token and a model name and nothing else, so an OpenRouter-style setup could not pass its routing headers or its provider and quantization options.

- `models.OpenAI` gains `ExtraHeaders` and `ExtraBody`, applied in `clientOptions()` through `option.WithHeader` and `option.WithJSONSet`. `ExtraBody` keys may be dotted paths, so `provider.order` writes a nested field.
- The fields are threaded through the config path — the API request structs, the stored model, the task params and the agent's scan request — so a model configured over the API or in JSON actually uses them.
- `pkg/database` stores them with `serializer:json`, and `AutoMigrate` adds the two columns.

## Type of change

- [x] New feature

## How was this tested?

- `go build ./...` passes.
- New tests in `common/utils/models` start a real HTTP server, point the client at it, and assert that both the headers and the nested `provider.order` field arrive, and that nothing extra is sent when the fields are unset.
- `go test ./common/utils/models/... ./pkg/database/...` passes.

Not in this PR, happy to follow up:

- the UI form field; the API and JSON path works today.
- the Python subprocess path (`common/agent/prompt_tasks.go` passes `--base_url` / `--api_key` / `--model` to AIG-PromptSecurity), which would need matching CLI flags on that side.

Two pre-existing failures I ran into while testing, both reproducing on a clean `f25e325` and unrelated to this change:

- `common/agent` cannot build its tests: `tasks_test.go` refers to undefined `TestDemoAgent`, `BaseUrl` and `Token`.
- `common/websocket` fails several tests that assert Chinese response messages.

## Related issue

Refs #85

## Checklist

- [x] `go build ./...` succeeds locally


boy-hack:
Thanks for picking up #85's backend half. The threading through the config path (API → stored model → task params → scan request → `OpenAI.clientOptions`) looks complete, and the test that spins up a real `httptest` server and asserts both headers and the nested `provider.order` field arrive is exactly the right kind of test. A few points before merge:

- **Security review needed for `extra_headers` storage.** `extra_headers` is persisted as JSON in the `models` table (and now travels through `ModelInfo` / `UpdateModelInfo`). Those headers can carry sensitive values (e.g. an `Authorization` / routing token). Please double-check that model config isn't dumped verbatim into any request/audit logs, and that any GET-model endpoint redacts these fields the same way the existing `token` field presumably is. `extra_headers` / `extra_body` need the same treatment.
- **Clearing the fields on update isn't currently possible.** `HandleUpdateModel` only sets `extra_headers` / `extra_body` when the incoming value is non-nil. Because the JSON tags are `omitempty`, a client can't *remove* headers/body that were previously set (sending an empty map is dropped). If that's intended, fine; otherwise consider an explicit "clear" path or accepting empty as a reset.
- **Verify `option.WithJSONSet` is in the pinned openai-go SDK version.** It's a relatively newer option; the test passing locally implies it is, but please confirm the `go.mod` floor actually includes it so CI (which may resolve a different patch) doesn't break.
- **Python subprocess path** (`prompt_tasks.go`) is correctly called out as future work — good that it's flagged rather than silently broken.

The two pre-existing `common/agent` / `common/websocket` test failures you noted reproduce on a clean `main`, so they're not regressions from this PR.

Solid contribution. Once the secret-logging question is confirmed, this is good to merge.

helo060228:
Thanks — point (2) was right in spirit, and chasing it found a real bug in the update path.

**1. Secret handling.** No log call prints the model config: the ones in `model_api.go` and `task_manager.go` log trace IDs, model IDs, names, counts and errors only. But the list and detail responses did not return the new fields at all, so the API was write-only for them. They now return them with keys visible and values masked, the same way `token` is (`maskValues`), covered by `common/websocket/model_masking_test.go`.

One narrow path I left alone because it is not specific to this PR: the DB is opened with `&gorm.Config{}`, so GORM's default logger interpolates parameters when it logs an error or a slow query. `ParameterizedQueries: true` closes that, and I am happy to do it separately.

**2. Clearing was broken, but not because of `omitempty`.** `omitempty` only affects marshalling, so `{"extra_headers": {}}` does arrive as a non-nil empty map. The real bug was in `UpdateModel`: a map value passed through `Updates(map)` skips the field serializer, and the driver rejected it with `sql: converting argument $1 type: unsupported type map[string]interface {}` — the UPDATE wrote nothing at all. Both fields now go through a struct with `Select`, and `pkg/database/model_extra_test.go` locks in store, clear and keep.

**3. Confirmed.** `go.mod` requires `github.com/openai/openai-go v1.8.3`, and `go list -m` resolves exactly v1.8.3, whose `option` package has `WithJSONSet` and `WithHeader`. The version is pinned, so CI cannot pick a different patch.

(4) agreed, the Python path stays a follow-up.

boy-hack:
@helo060228 Thanks for the detailed follow-up — I pulled the updated diff and verified each point:

1. **Secret handling — looks correct now.** `maskValues` returns only the *keys* with values replaced by `maskedToken`, and it's applied on both `HandleGetModelList` and `HandleGetModelDetail`. Combined with the write-path tests in `model_masking_test.go`, the `extra_headers` / `extra_body` values no longer leak to clients. Good.
2. **The update bug was real and is fixed properly.** Your diagnosis is exactly right: a `map` value routed through `Updates(map[string]interface{})` skips the field `serializer:json` and the SQLite driver rejects it with `unsupported type map[string]interface {}`. Routing these two fields through a struct + `Select` (now in `UpdateModel`) both fixes the write *and* makes an empty `{}` a real "clear" — which also resolves the original "can't unset" concern from my first review. `model_extra_test.go` locking in store/clear/keep is the right coverage.
3. **SDK pin confirmed.** `go.mod` floors `openai-go v1.8.3`, `WithJSONSet`/`WithHeader` both exist there, so CI can't resolve a different patch. Solid.

One thing I'd like to keep on the radar before this ships: **the GORM logger interpolation gap you flagged.** You're right that it's narrow and not specific to this PR, but it *is* the same secret-handling threat model I raised in (1) — `&gorm.Config{}` (in `pkg/database/config.go`) leaves GORM's default logger on, which interpolates bound parameters into error/slow-query logs. Now that `extra_headers`/`extra_body` values are persisted in the DB, a single failed/aborted `UPDATE` or a slow query could print those exact secret values into logs. I'd strongly suggest landing `ParameterizedQueries: true` (or a `Logger` that doesn't interpolate) in the *same* release window as this PR, even if it's a separate commit/PR. It's a 1-line change and closes the loop on the whole secret-exposure concern rather than leaving a known leak path open next to a feature that stores secrets.

Everything else (the `sortedKeys` stability for tests, the Python subprocess path explicitly deferred) is fine.

Net: this is in great shape. The only blocker for my "good to merge" is agreeing on a plan for the GORM logging — if you'd rather track it separately, please open the issue/PR and link it here so reviewers know it's covered. Don't merge until that's resolved.

helo060228:
Landed the GORM logger change here rather than tracking it separately, since you are right that it is the same threat model. `pkg/database/config.go` now builds the logger explicitly with `ParameterizedQueries: true` (Warn level, 1s slow threshold, same writer as the default).

Verified by forcing a failing query through the same `InitDB`:

```
SQL logic error: no such table: definitely_missing_table (1)
[0.528ms] [rows:0] SELECT * FROM definitely_missing_table WHERE secret = ?
```

The bound value no longer appears anywhere in the output, where before this change the same path printed the interpolated parameters. `go build ./...` and `go test ./pkg/database/...` pass.

That closes (1) from your first review, so I think this one is ready whenever you want to merge it.

boy-hack:
@helo060228 Verified the GORM logger change in `pkg/database/config.go` — `ParameterizedQueries: true` closes the parameter-interpolation leak, and your forced-failure query shows the bound value no longer appears in the log line (`SELECT * FROM definitely_missing_table WHERE secret = ?`). That was the last open item from my original review (secret handling on the persist path).

With `maskValues` applied on both `HandleGetModelList` and `HandleGetModelDetail`, and the `UpdateModel` store fix (routing the two map fields through a struct + `Select`, which correctly handles the `serializer:json` skip *and* turns an empty `{}` into a real clear), the whole secret-exposure concern is now closed. The `extra_headers`/`extra_body` threading and the `option.WithJSONSet`/`WithHeader` `httptest` test are solid.

This is ready to merge from my side. One tiny non-blocking follow-up thought: the new logger uses stdlib `log` to stdout rather than the project's `gologger`, but that matches the previous GORM default behavior, so it's fine as-is.

helo060228:
Thanks for the careful pass on both halves.

On the logger: keeping the stdlib `log` writer was deliberate — it is the same writer the previous default used, so the only behavioural change is the parameterized output. If you would rather have these lines go through `gologger` for consistency with the rest of the project, that is a two-line change and I am happy to add it here or as a follow-up, whichever you prefer.

PR #671, opened 2026-09-21
fix(deepteam): RailFence crashes with IndexError when rails <= 1

Linxiushen:
## Problem

`AIG-PromptSecurity/deepteam/attacks/single_turn/encoding/rail_fence.py` simulates the fence with a list of `rails` rows and bounces the cursor at `rail == 0` / `rail == self.rails - 1`. With `rails=1` both edges are index 0, so the cursor steps to 1 and never returns; the next `rails[rail].append(char)` raises `IndexError`. With `rails<=0` the row list is empty and the first append fails the same way:

```
rails=3 'HELLOWORLD' -> 'HOLELWRDLO'
rails=2 'HELLOWORLD' -> 'HLOOLELWRD'
rails=1 'HELLOWORLD' -> IndexError: list index out of range
rails=0 'a'          -> IndexError: list index out of range
```

`rails` is not an internal detail: `python cli_run.py --scan-tools techniques` (via `ToolScanner._extract_parameters`) lists it as an optional parameter with default 3 and no range, and `cli/parsers.py::parse_attack` passes `--techniques "RailFence:rails=1"` straight into the constructor. What happens next depends on the entry point:

- Python API `red_team(..., ignore_errors=False)` (the default): the `IndexError` propagates and the whole run aborts.
- CLI (`cli/red_team_runner.py` uses `ignore_errors=True`): the run continues, every RailFence case is marked as an error, and the report attributes it to "The attack simulation phase failed, possibly due to rate limiting or security blocking by the generalization model" — a misleading explanation for an encoder crash.

## Fix

- `rails == 1`: return the input unchanged. A single-rail rail-fence cipher is the identity, and the framework already treats an unchanged enhancement as a first-class outcome (`attack_simulator.py` marks it `useless`).
- `rails < 1`: raise `ValueError` with a message naming the parameter, following the convention `base_encoding.py` / `rot_encoding.py` use for invalid options, instead of an `IndexError` from inside the loop.
- `rails >= 2`: untouched. Verified by diffing the original and patched `enhance()` over rails 2..32 × 3006 inputs (empty, single char, long, mixed CJK/emoji/control chars): 93,186 comparisons, 0 mismatches.

## Test

`AIG-PromptSecurity/tests/test_rail_fence.py`: existing outputs for rails 2/3/5 preserved, `rails=1` is the identity, `rails in (0, -3)` raise `ValueError`. On `main`: 3 failed / 3 passed; with this change: 6 passed. The rest of `tests/` (excluding `test_custom_vulnerability.py`, which calls a live model at import time) still passes: 40 passed.

Run: `cd AIG-PromptSecurity && pytest tests/test_rail_fence.py -v`

_Written with Claude Code (AI-assisted) and submitted under the account owner's authorization._


boy-hack:
Thanks for the thorough fix and the detailed write-up. The approach is sound and the analysis of the two failure modes (rails=1 bounce-degeneration, rails<=0 empty row list) is exactly right. A few notes:

- **The `rails == 1` identity return is the correct call.** Returning the input unchanged is the right contract — a single-rail fence is the identity, and letting the framework mark it `useless` is the intended outcome rather than crashing the run or emitting a misleading "rate limiting" error in the CLI path.
- **`rails < 1` raising `ValueError` is much better than the old `IndexError`.** Naming the parameter in the message aids debugging.
- **Test coverage is strong** — existing outputs preserved (rails 2/3/5), identity for rails=1, `ValueError` for 0/-3, plus the 93k-comparison diff check you mention. Good regression guard.

One small nit: the `ValueError` text is in Chinese (`不支持的栅栏数: ...`). I see you're matching the convention in `base_encoding.py` / `rot_encoding.py`, which is fine for internal consistency, but since `RailFence(rails=...)` is a public API surface a non-Chinese-speaking consumer could hit, an English message (or bilingual) would be friendlier. Not blocking — your call.

Also worth a thought: other `encoding` attacks may have similar edge cases (empty/edge inputs), but that's out of scope here.

LGTM overall. Nice work.

Linxiushen:
Thanks for the thorough review and the kind words. On the one nit: you're right that `RailFence(rails=...)` is a public surface, so I made the `rails < 1` message bilingual — an English sentence first, then the original Chinese, keeping consistency with `base_encoding.py` / `rot_encoding.py`:

```
Unsupported number of rails: {rails}. rails must be an integer >= 1. 不支持的栅栏数: {rails}，rails 必须是大于等于 1 的整数
```

The string still contains `rails`, so the existing `pytest.raises(ValueError, match="rails")` assertion is unaffected. I verified in isolation that the existing encodings (rails 2/3/5) are byte-identical, rails=1 stays the identity, and rails 0/-3 raise the new bilingual `ValueError`.

Good call on the other encodings possibly having similar edge cases — agreed it's out of scope here, but I'm happy to open a separate follow-up sweeping the `encoding/` attacks for empty/degenerate inputs if that's useful.

boy-hack:
Verified the fix logic locally (loaded the patched `rail_fence.py` with stubbed deps):

| Case | main | this PR |
|---|---|---|
| `rails=3` 'HELLOWORLD' | `HOLELWRDLO` ✓ | `HOLELWRDLO` ✓ (unchanged) |
| `rails=2` 'HELLOWORLD' | `HLOOLELWRD` ✓ | `HLOOLELWRD` ✓ (unchanged) |
| `rails=1` | `IndexError` | identity transform, no crash ✓ |
| `rails=0` / negative | `IndexError` | `ValueError` with clear bilingual message ✓ |

The root-cause analysis is accurate: with `rails=1` the bounce condition degenerates (`rail == 0` and `rail == self.rails - 1` become the same edge), the cursor escapes and never returns. Guard + identity return is the right minimal fix, and adding regression tests (`tests/test_rail_fence.py`) is appreciated.

Nit from review: the new `ValueError` message mixes English and Chinese. AIG's convention for this repo is English-only runtime messages in `AIG-PromptSecurity` (README/code English-first; see CONTRIBUTING). Suggest dropping the Chinese sentence or splitting into a localized warning upstream. Non-blocking.

Linxiushen:
Thanks for the careful review and the local verification table — much appreciated.

Addressed the message nit in 3b85b09: the `ValueError` for `rails < 1` is now English-only, matching the English-first runtime-message convention in `AIG-PromptSecurity`. The message still names `rails` so the regression test's `match="rails"` holds, and behaviour is otherwise unchanged.

Re-verified the same cases after the change (isolated load of the patched `rail_fence.py`):
- rails=3 'HELLOWORLD' -> 'HOLELWRDLO' (unchanged)
- rails=2 'HELLOWORLD' -> 'HLOOLELWRD' (unchanged)
- rails=1 -> identity, no crash
- rails=0 / -3 -> `ValueError: Unsupported number of rails: 0. rails must be an integer >= 1.`

Good point on the other encoding attacks possibly having similar edge cases — happy to open a separate follow-up for those if you'd like.

(For disclosure: this PR was prepared with AI assistance and verified locally by me.)

boy-hack:
@Linxiushen Verified the change in `3b85b09` — the `rails < 1` `ValueError` is now English-only, matching the `AIG-PromptSecurity` English-first runtime-message convention, and the `match="rails"` regression test still holds since the message names `rails`. Combined with the `rails == 1` identity return and the new `tests/test_rail_fence.py` (preserved rails 2/3/5 outputs, identity for 1, `ValueError` for 0/-3), this is a clean, well-tested fix.

LGTM — ready to merge from my side. Thanks for the quick turnaround on the message nit, and for the offer to sweep the other `encoding/` attacks for edge cases; opening a follow-up issue/PR for that would be a good idea.

